PR #6, opened 2024-01-12
created scaffold for started recipes db

CodeSmitty:
## Description

https://trello.com/c/OwcKgWQW/3-create-a-basic-data-model-for-recipe-table

Changed project db from sqlite to psql. Altered Gemfile, replaced  sqlite gem with pg. 

To switch application to use psql and update database.yml file I ran.
`rails db:system:change --to=postgresql`
 
Restarted psql server
`sudo service postgresql restart`

Went into psql user console
`psql -d mame_zefu_development -U smitty -W`

Switched connection to mame_zefu_database
`\c`

Generated Scaffold for Recipes
`rails generate scaffold Recipes name:string ingredients:text directions:text`

Finished with `rails db:migrate`

I tested application in local server. I am able to create, update, destroy recipes. As well as in rails console. 

## Testing

I struggled with getting psql to work, so had to research a bit with psql commands. Once I was able to create a database and assign a user to the correct port. I was able to run migrations and test forms. I created a couple recipes, edited both and destroyed one. Changes reflected when I tested in the rails console for `Recipe.all`


nfoos:
@CodeSmitty Thanks for taking the time to fill out the Description and Testing sections of the PR ❤️ 

Did you make any manual changes to the controller or view files? I suspect that the only manual edit you made with this change is the database user, is that correct?

config/database.yml
```diff
@@ -1,25 +1,87 @@
...
 development:
   <<: *default
-  database: db/development.sqlite3
+  database: mame_zefu_development
+
+  # The specified database role being used to connect to postgres.
+  # To create additional roles in postgres see `$ createuser --help`.
+  # When left blank, postgres will use the default role. This is
+  # the same name as the operating system user running Rails.
+  username: smitty
```
nfoos: I suggest using a generic username, not one specific to you.  You should be able to create a `mame_zefu` user in your development environment.

nfoos (commented):
One addition I'd like to see is requiring the `name` field on recipes. This should involve a validation in the model and a new database migration.

I'm surprised the scaffold didn't add any tests. When I messed with the generators the other day for models and controllers it added a simple controller test.

Assuming you add the validation I suggested, I encourage you to add a simple test for it. https://www.codewithjason.com/write-model-tests-part/ is a pretty good example of adding a simple test on the model checking that it validates the presence of a field.  This will require adding the rspec gem (the only way I'm really familiar with testing anyway)

config/database.yml
```diff
@@ -1,25 +1,87 @@
...
 #
 default: &default
-  adapter: sqlite3
+  adapter: postgresql
+  encoding: unicode
+  host: localhost
+  # For details on connection pooling, see Rails configuration guide
+  # https://guides.rubyonrails.org/configuring.html#database-pooling
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  timeout: 5000
```
nfoos, line 26: The username and password can be set in this default section instead, that way it applies to both the development and test environments.

```suggestion
  username: mame_zefu
  password: password
```

config/database.yml
```diff
@@ -1,25 +1,87 @@
...
+#
+# Read https://guides.rubyonrails.org/configuring.html#configuring-a-database
+# for a full overview on how database connection configuration can be specified.
+#
 production:
   <<: *default
-  database: db/production.sqlite3
+  database: mame_zefu_production
+  username: mame_zefu
+  password: <%= ENV["MAME_ZEFU_DATABASE_PASSWORD"] %>
```
nfoos: These settings aren't necessary. Instead you can specify all the connection details with the DATABASE_URL environment variable

CodeSmitty: Ok, I'll look into it and see if I can set it up. Along with the other additions/changes. I know on mac (from the course). When you switch from sql to psql it automatically adds the default url. On windows/was you had to do more set up if you wanted it that way I believe. I do these changes on this s!same branch correct?

nfoos: Yes, make the changes on the same branch. I suggest doing the config changes in one commit and the validation/test in another.

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
+  resources :recipes
   # Define your application routes per the DSL in https://guides.rubyonrails.org/routing.html
 
   # Defines the root path route ("/")
```
nfoos: Should set the root route to the recipe index

nfoos (changes requested):

.gitignore
```diff
@@ -36,3 +36,4 @@
 
 /app/assets/builds/*
 !/app/assets/builds/.keep
+/DATABASE_URL.env
```
nfoos: Should remove this and the local file if it exists

app/views/layouts/application.html.erb
```diff
@@ -11,7 +11,7 @@
     <%= javascript_importmap_tags %>
   </head>
 
-  <body>
+  <body >
```
nfoos: Remove whitespace
```suggestion
  <body>
```

nfoos (changes requested):

app/models/recipe.rb
```diff
@@ -0,0 +1,3 @@
+class Recipe < ApplicationRecord
+    validates :name, length:{minimum:3, maximum:25}, presense:true
```
nfoos: Should be `presence`, not `presense`

nfoos: Length of 25 is probably too restrictive. Honestly, I'm not sure there's much value in validating the length

nfoos: It would also be good to generate a database migration to enforce that name is required.

nfoos (changes requested):

app/models/recipe.rb
```diff
@@ -0,0 +1,3 @@
+class Recipe < ApplicationRecord
+    validates :name, presence:true
```
nfoos: Should put a space after the colon
```suggestion
    validates :name, presence: true
```

spec/models/recipes.rb
nfoos, line 1: This should be named `recipes_spec.rb` for rspec to detect and run it

app/models/recipe.rb
```diff
@@ -0,0 +1,3 @@
+class Recipe < ApplicationRecord
+    validates :name, presence: true
```
nfoos:
```suggestion
  validates :name, presence: true
```

nfoos (approved):
Looks good 🚀

PR #8, opened 2024-01-16
Cleaned up UI in index, show, edit, form, new. Added navigation

CodeSmitty:
Organized the UI. 

Changed button naming on the form to something more clean & ui friendly. 
I removed some buttons from the show page. 

Removed Recipe ingredients/directions from the root page. Made Recipe name the title and link to the show page.  

Recipe show page, added simple format for line breaks in the ingredients & directions text area. 

I created a basic nav bar, using Taillwind components. There is a problem with the Javascript that will still need to be fixed. Specifically mobile hamburger button.  This navbar requires Flowbite js, I need to see if it's compatible with what we are working with. If not, I will need to look at another option. Stimulus Js seems to be a good option.  I don't think a seperate ticked will be needed for this. 

Added flash message code in the show.html.erb file for future flash messages. This can be removed, since the original is working correctly. Might be something for discussion if we want to use flash messages. Added a turbo-comfirm argument to the delete button in the same file. This prompts an alert `Are you sure you want to delete this recipe?`. 

## Description

These changes were required to make the ui easier to navigate for the user. Creating simple buttons, layouts, makes it much more user friendly. 

## Testing

I created, updated, destroyed recipes to test the buttons and links all worked as intended. The navigation bar works well on desktop. I destroyed recipes to test the alert prompt, which worked as intended.


app/views/recipes/show.html.erb
```diff
@@ -3,13 +3,18 @@
...
+      <% end %>
 
     <%= render @recipe %>
 
-    <%= link_to "Edit this recipe", edit_recipe_path(@recipe), class: "mt-2 rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
+    <%= link_to "Edit", edit_recipe_path(@recipe), class: "mt-2 rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
     <div class="inline-block ml-2">
-      <%= button_to "Destroy this recipe", recipe_path(@recipe), method: :delete, class: "mt-2 rounded-lg py-3 px-5 bg-gray-100 font-medium" %>
+      
+      <%= button_to "Delete", recipe_path(@recipe), method: :delete,data:{ turbo_confirm: "Are you sure?" }, class: "mt-2 rounded-lg py-3 px-5 bg-gray-100 font-medium" %>
```
nfoos: Formatting
```suggestion
      <%= button_to "Delete", recipe_path(@recipe), method: :delete, data: { turbo_confirm: "Are you sure?" }, class: "mt-2 rounded-lg py-3 px-5 bg-gray-100 font-medium" %>
```

app/views/recipes/show.html.erb
```diff
@@ -3,13 +3,18 @@
     <% if notice.present? %>
       <p class="py-2 px-3 bg-green-50 mb-5 text-green-500 font-medium rounded-lg inline-block" id="notice"><%= notice %></p>
     <% end %>
+    <% flash.each do |type, msg| %>
+        <div>
+          <%= msg %>
+        </div>
+      <% end %>
```
nfoos: Formatting
```suggestion
    <% flash.each do |type, msg| %>
      <div>
        <%= msg %>
      </div>
    <% end %>
```

app/views/recipes/index.html.erb
```diff
@@ -3,12 +3,13 @@
...
+  <div class="flex justify-center items-center">
     <%= link_to "New recipe", new_recipe_path, class: "rounded-lg py-3 px-5 bg-blue-600 text-white block font-medium" %>
   </div>
 
-  <div id="recipes" class="min-w-full">
-    <%= render @recipes %>
+  <div id="recipes" class="min-w-full flex-column">
+    <% @recipes.each do |recipe|%>
+        <%= link_to "#{recipe.name}", recipe, class: "block font-medium mb-1 text-2xl p-4 text-center" %>
+    <%end%>
```
nfoos: Formatting
```suggestion
  <div id="recipes" class="min-w-full flex-column">
    <% @recipes.each do |recipe| %>
      <%= link_to "#{recipe.name}", recipe, class: "block font-medium mb-1 text-2xl p-4 text-center" %>
    <% end %>
```

app/views/recipes/index.html.erb
```diff
@@ -3,12 +3,13 @@
     <p class="py-2 px-3 bg-green-50 mb-5 text-green-500 font-medium rounded-lg inline-block" id="notice"><%= notice %></p>
   <% end %>
 
-  <div class="flex justify-between items-center">
-    <h1 class="font-bold text-4xl">Recipes</h1>
+  <div class="flex justify-center items-center">
     <%= link_to "New recipe", new_recipe_path, class: "rounded-lg py-3 px-5 bg-blue-600 text-white block font-medium" %>
```
nfoos: This button is a bit redundant with the 'Create Recipe' link in the navbar

app/assets/images/One_piece_logo.png
nfoos, line 1: Nice image, but what if the logo was a pinto bean with a mustache wearing a chef's hat?

image

package.json
nfoos, line 1: With this change adding one more command that needs to be ran when deploying, it might be a good idea to create a script that can be called.

Call the script something like `bin/build` and make it executable.
```
#!/usr/bin/env bash
# exit on error
set -o errexit

bundle install
# install JavaScript dependencies here
bundle exec rake assets:precompile
bundle exec rake assets:clean
bundle exec rake db:migrate
```

nfoos: I was also looking at the available rake tasks included in the rails app using `bundle exec rake -T`

I see that there is a `yarn:install` task which suggests some preference of yarn over npm.  I'm certainly no expert when it comes to JavaScript package management, but I wonder if we shouldn't be using yarn instead.

app/views/layouts/application.html.erb
```diff
@@ -6,11 +6,10 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
     <%= stylesheet_link_tag "tailwind", "inter-font", "data-turbo-track": "reload" %>
-
     <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
     <%= javascript_importmap_tags %>
   </head>
-
+    <%= render "layouts/navigation"%>
```
nfoos, line 13: This belongs in the `<body>` and possibly within a `<header>`.  For reference, I was looking at this project that I worked on as part a rails tutorial: https://github.com/nfoos/rails-microblog/blob/master/app/views/layouts/application.html.erb

app/views/recipes/edit.html.erb
```diff
@@ -3,6 +3,5 @@
 
   <%= render "form", recipe: @recipe %>
 
-  <%= link_to "Show this recipe", @recipe, class: "ml-2 rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
-  <%= link_to "Back to recipes", recipes_path, class: "ml-2 rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
+  <%= link_to "Cancel", @recipe, class: "ml-2 rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
```
nfoos, line 6: I think the 'Back to recipes' button in `new.html.erb` could be changed to 'Cancel' as well

app/views/layouts/_navigation.html.erb
```diff
@@ -0,0 +1,26 @@
+
+
+<nav class="bg-white border-gray-200 dark:bg-gray-900">
+  <div class="max-w-screen-xl flex flex-wrap items-center justify-between mx-auto p-4">
+    <div class="flex items-center space-x-3 rtl:space-x-reverse">
+        <%= link_to image_tag("One_piece_logo.png"), recipes_path, class: "h-8 mt-1 logo-nav" %>
+        <span class="self-center text-2xl font-semibold whitespace-nowrap dark:text-white">Mame-Zefu</span>
```
nfoos: image

I feel like the text should be either as tall as the logo or centered vertically and maybe closer to the logo.  It would also be nice if it was part of the link. Not sure how I feel about the hyphen, maybe try out 'Mame Zefu' and 'MameZefu' and decide which of the 3 you like better.

app/views/layouts/_navigation.html.erb
```diff
@@ -0,0 +1,26 @@
...
+        <span class="sr-only">Open main menu</span>
+        <svg class="w-5 h-5" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 17 14">
+            <path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M1 1h15M1 7h15M1 13h15"/>
+        </svg>
+    </button>
+    <div class="hidden w-full md:block md:w-auto" id="navbar-default">
+      <ul class="font-medium flex flex-col p-4 md:p-0 mt-4 border border-gray-100 rounded-lg bg-gray-50 md:flex-row md:space-x-8 rtl:space-x-reverse md:mt-0 md:border-0 md:bg-white dark:bg-gray-800 md:dark:bg-gray-900 dark:border-gray-700">
+        <li>
+          <%= link_to "Home", recipes_path, class: "block py-2 px-3 text-white bg-blue-700 rounded md:bg-transparent md:text-blue-700 md:p-0 dark:text-white md:dark:text-blue-500"%>
```
nfoos: Not sure how useful the Home link is if you know that you can click on the logo instead 🤷‍♂️

app/views/layouts/_navigation.html.erb
```diff
@@ -0,0 +1,26 @@
...
+        </li>
+        <li>
+          <%= link_to "Create Recipe", new_recipe_path, class:"block py-2 px-3 text-gray-900 rounded hover:bg-gray-100 md:hover:bg-transparent md:border-0 md:hover:text-blue-700 md:p-0 dark:text-white md:dark:hover:text-blue-500 dark:hover:bg-gray-700 dark:hover:text-white md:dark:hover:bg-transparent" %>
```
nfoos: I'm on the fence about this. For currently functionality it makes sense but maybe I'd change the language to 'Add Recipe' or 'New Recipe'. If the future of the app is to do more than just create and edit recipes, I'd expect the navbar links to be more broad like 'Recipes', 'Lists', 'Plan', etc.

app/views/layouts/_navigation.html.erb
```diff
@@ -0,0 +1,26 @@
+
+
+<nav class="bg-white border-gray-200 dark:bg-gray-900">
```
nfoos: I don't recall exactly how to do it, but how would you feel about having the navbar always visible even when scrolling? If that were to be done, I think the navbar would need to have a different background color than the main content otherwise it might look weird.

nfoos (commented):
Some formatting issues and a bunch of minor suggestions that are open for discussion and don't necessarily need to be acted on.

app/views/recipes/show.html.erb
CodeSmitty: I believe I removed this line on my most recent commit. Since I removed the flash message in the controller.

app/views/recipes/index.html.erb
CodeSmitty: Yea, I didn't put a lot of thought on the navigation links. I based it on what most websites have, I just wanted to get it to work. I can remove the home link, and leave the create recipe link. I'll remove the button as well. Right now since we don't have a lot of routes, I'm unsure what other links to add to nav bar. I'm sure at some point we will have a login/out link, maybe user profile etc. Unless you have some ideas we can add right now.

app/views/layouts/_navigation.html.erb
CodeSmitty: I'm ok with that. I believe it's the fixed position in css.

app/views/layouts/_navigation.html.erb
CodeSmitty: Yea, I can mess with the alignment, it will be better once I get the logo. I was probably do that tonight, and have a logo made in the next couple of days. I honestly like 'MameZefu' the most.

package.json
CodeSmitty: I myself have not used yarn very much, so I wasn't sure. Good to know on the rake command, I wasn't aware of that one. The main reason I used npm was based on the docs for flowbite js. I do know a little about yarn, but I have not really used it. So I didn't think about it already existing in the application. Would it be too much to remove the package/node modules and re install with yarn? I will have to look into creating a script to run the build commands. I haven't done something like that yet.

app/views/recipes/index.html.erb
nfoos: Nope no ideas for anything at the moment which I think is fine

package.json
nfoos: Yeah, don't worry about yarn.

For the script, basically just create a file with the contents I provided, substitute in the npm install where the comment is and make it executable before committing. Then your render build command can just be the new script name

app/views/layouts/_navigation.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  <div class="max-w-screen-xl flex flex-wrap items-center justify-between mx-auto p-4">
+    <div class="flex items-center space-x-3 rtl:space-x-reverse">
+        <%= link_to image_tag("One_piece_logo.png"), recipes_path, class: "h-8 mt-1 logo-nav" %>
```
nfoos: I know they're essentially the same, but this should link to the root path
```suggestion
        <%= link_to image_tag("One_piece_logo.png"), root_path, class: "h-8 mt-1 logo-nav" %>
```

nfoos (approved):
**LGTM** :shipit:

PR #9, opened 2024-01-24
updated migration/table  to include new fields. Reflected in the reci…

CodeSmitty:
…pe form, show,and edit pages

Updated db tables & migrations with  `:yield, :prep_time, :cook_time, :description, :rating, :is_favorite, :notes` fields. 
Updated form, edit, show pages to reflect new additions to the db table. 

## Description

https://trello.com/c/V5VQV0WI/7-add-fields-to-recipes

`rails g AddNewFieldsToRecipes`

In the new migration I added a column for each of these fields. `:yield, :prep_time, :cook_time, :description, :rating, :is_favorite, :notes` fields. 

set `is_favorite, default: false & null: false`

In psql is altered table -> column is_favorite to not null. `ALTER TABLE recipes ALTER COLUMN is_favorite SET NOT NULL;`

Added new fields in the `_form.html.erb helper` 

Updated the edit and show page  with the new table additions. 

Generated a recipes seed for quicker testing. 

## Testing

I first checked the database tables to make sure the new migrations were showing as intended. 
I then generated a seed and destroyed all current data and injected the db with the new seed data.
Show page rendered all new fields correctly. 
I clicked on random recipes and clicked on edit. I changed each field and saved one by one to make sure all them updated correctly. 

I manually created recipes  testing each field one at a time.  All data worked correctly and all changes were reflected in the db.


app/views/recipes/_recipe.html.erb
```diff
@@ -3,6 +3,35 @@
...
+  </p>
+
+  <p class="my-5">
+    <strong class="block font-medium mb-1"> <%= recipe.is_favorite == true ?   "Favorite: yes"  : " " %></strong>
+  </p>
+
+  <p class="my-5">
+    <strong class="block font-medium mb-1">Description:</strong>
+    <%= simple_format(recipe.description) %>
+  </p>
```
nfoos: @CodeSmitty I think it would be nice if the fields that are empty were hidden. In other words, if a recipe doesn't have a description, I don't want to see the "Description:" label with no contents. What do you think?

app/views/recipes/_recipe.html.erb
```diff
@@ -3,6 +3,35 @@
...
+    <%= simple_format(recipe.cook_time) %>
+  </p>
+
+  <p class="my-5">
+    <strong class="block font-medium mb-1">Rating:</strong>
+    <%= recipe.rating %>
+  </p>
+
+  <p class="my-5">
+    <strong class="block font-medium mb-1"> <%= recipe.is_favorite == true ?   "Favorite: yes"  : " " %></strong>
```
nfoos: Minor formatting, but also the explicit check against `true` shouldn't be necessary, the truthiness of `is_favorite` evaluates the same
```suggestion
    <strong class="block font-medium mb-1"> <%= recipe.is_favorite ? "Favorite: yes" : "" %></strong>
```

app/views/recipes/_recipe.html.erb
```diff
@@ -3,6 +3,35 @@
     <%= recipe.name %>
   </h1>
 
+  <p class="my-5">
+    <strong class="block font-medium mb-1">Yield:</strong>
+    <%= simple_format(recipe.yield) %>
```
nfoos: I'm not sure that `simple_format` is necessary for the string fields, but it probably doesn't hurt?

app/views/recipes/_form.html.erb
```diff
@@ -16,6 +16,38 @@
...
+    </div>
+  </div>
+
+  <div class="my-5">
+    <%= form.label :rating %>
+    <%= form.number_field :rating, rows: 4, class: "block shadow rounded-md border border-gray-200 outline-none px-3 py-2 mt-2 w-2/12" %>
+  </div>
+  
+  <div class="my-5">
+    <%= form.label :is_favorite %>
```
nfoos: Rather than the label being `Is favorite []` I think it might be better as `Favorite? []` and I think the checkbox itself should be to the right of the label.

app/views/recipes/_form.html.erb
```diff
@@ -16,6 +16,38 @@
...
+
+    <div>
+      <%= form.label :cook_time %>
+      <%= form.text_field :cook_time, class: "block shadow rounded-md border border-gray-200 outline-none px-3 py-2 mt-2 w-full" %>
+    </div>
+  </div>
+
+  <div class="my-5">
+    <%= form.label :rating %>
+    <%= form.number_field :rating, rows: 4, class: "block shadow rounded-md border border-gray-200 outline-none px-3 py-2 mt-2 w-2/12" %>
```
nfoos: It would be cool if the input for this was 5 stars that you could click ⭐ ⭐ ⭐ ⭐ ⭐ 

If that's too much work, I think the input box should be smaller and can be positioned to the right of the label.

app/views/recipes/_form.html.erb
```diff
@@ -16,6 +16,38 @@
...
+  </div>
+
+  <div class="my-5">
+    <%= form.label :description %>
+    <%= form.text_area :description, rows: 4, class: "block shadow rounded-md border border-gray-200 outline-none px-3 py-2 mt-2 w-full" %>
+  </div>
+
   <div class="my-5">
     <%= form.label :ingredients %>
     <%= form.text_area :ingredients, rows: 4, class: "block shadow rounded-md border border-gray-200 outline-none px-3 py-2 mt-2 w-full" %>
```
nfoos: I know that `ingredients` and `directions` weren't touched in this change, but what do you think about making their boxes a little taller to reduce the need for scrolling or manual resizing?

db/seeds.rb
```diff
@@ -5,3 +5,183 @@
...
+  {
+    name: 'Spaghetti Bolognese',
+    yield: '4 servings',
+    ingredients: "Ground beef,\nOnion,\nGarlic,\nTomato sauce,\nSpaghetti,\nOlive oil",
+    directions: "1. Cook spaghetti according to package instructions.\n2. Brown ground beef with onion and garlic.\n3. Add tomato sauce and simmer.\n4. Serve sauce over cooked spaghetti.",
+    prep_time: 15,
+    cook_time: 30,
+    description: 'Classic Italian dish with savory meat sauce over spaghetti.',
+    notes: 'You can add Parmesan cheese on top for extra flavor.',
+    rating: 4.5,
```
nfoos: Not a big deal, but these decimal ratings get truncated to integers

db/seeds.rb
```diff
@@ -5,3 +5,183 @@
...
+    description: 'Healthy and refreshing quinoa salad with Mediterranean flavors.',
+    notes: 'Add grilled chicken for a protein boost.',
+    rating: 4.2,
+    is_favorite: false
+  },
+]
+
+recipes_data.each do |recipe_data|
+  Recipe.create!(recipe_data)
+end
```
nfoos: I'd like to be able to run the seeds multiple times without getting errors or duplicate data. The following should find a recipe with the same name and update it or create a new one.

```suggestion
recipes_data.each do |recipe_data|
  Recipe.find_or_initialize_by(name: recipe_data[:name]).update(recipe_data)
  end
end
```

db/seeds.rb
nfoos, line 1: We wouldn't want these recipes added in production if the seeds are ran in that environment. There are various examples online for splitting the seeds files.  I think I personally prefer splitting them up by the table being populated and just having a statement to exit at the top depending on the environment.

For example, update `seeds.rb` to the following which will run each file in `db/seeds/` in alphanumeric order:
```
Dir[Rails.root.join('db/seeds/*.rb')].sort.each do |file|
  require file
end
```

Then you can have `db/seeds/001_recipes.rb`
```
exit unless Rails.env.development?

# insert recipe records here
```
As you add seeds files, you can adjust the number in the filename to control which order they run.

nfoos (commented):
Looks pretty good. I have some suggestions on the views and some changes I'd like made to the seeds.

app/views/recipes/_recipe.html.erb
CodeSmitty: Do you want that logic on every field or just like notes and description? I already have something like that on is_favorite.

app/views/recipes/_recipe.html.erb
CodeSmitty: Simple format is mostly usefull for the text fields. Like directions, description, notes. It wouldn't hurt on the name and other one line fields. I added it to help with the new line formatting. We can remove it on the fields it doesn't need to be in if you want.

app/views/recipes/_recipe.html.erb
nfoos: I think it would be useful on all fields except name which is required and maybe ingredients/directions which aren't required but a recipe is incomplete without

nfoos: One way to accomplish this is to wrap the section in an if-block:

```
<% if recipe.description.present? %>
  <p class="my-5">
    <strong class="block font-medium mb-1">Description:</strong>
    <%= simple_format(recipe.description) %>
  </p>
<% end %>
```

app/views/recipes/_recipe.html.erb
nfoos, line 1: These fields take up a lot of vertical space. I'd like to see as much of these on the same line as possible. Some ways to save horizontal space if they are too long are:
- shorten or exclude the label if it doesn't provide value
- use an icon in place of the label or value
image

db/schema.rb
```diff
@@ -10,16 +10,29 @@
...
+ActiveRecord::Schema[7.0].define(version: 2024_01_24_135559) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "categories", force: :cascade do |t|
+    t.string "name", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
```
nfoos: This should not have been committed
```suggestion
```

db/migrate/20240114021324_set_name_to_not_null.rb
```diff
@@ -1,5 +1,5 @@
 class SetNameToNotNull < ActiveRecord::Migration[7.0]
   def change
-    change_column_null :recipes, :name,  null: false
+    change_column_null :recipes, :name,  false
```
nfoos, line 3: Typically, you wouldn't want to modify migrations that have already ran in production, but I'll let this one slide

db/seeds.rb
```diff
@@ -5,3 +5,10 @@
 #
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+# db/seeds.rb
+
+# db/seeds.rb
+
```
nfoos: These lines are unnecessary. You could remove the comments above as well
```suggestion
```

nfoos (changes requested):

nfoos (approved):
**LGTM** 🚢

db/migrate/20240114021324_set_name_to_not_null.rb
CodeSmitty: So what is the move in those scenarios? I know you definitely don't want to do that when you already have many migrations.

PR #10, opened 2024-01-29
New categories

CodeSmitty:


db/seeds/002_categories.rb
```diff
@@ -0,0 +1,29 @@
+# db/seeds.rb
+exit unless Rails.env.development?
+
```
nfoos: We want this to run in production
```suggestion
```

db/seeds/002_categories.rb
```diff
@@ -0,0 +1,29 @@
...
+  { name: 'Seafood' },
+  { name: 'Poultry' },
+  { name: 'Beef' },
+  { name: 'Pork' },
+  { name: 'Sauce' },
+  { name: 'Condiment' }
+]
+
+categories_data.each do |category_data|
+  Category.find_or_create_by(name: category_data[:name]).update(category_data)
```
nfoos: The `update` is not needed here
```suggestion
  Category.find_or_create_by(name: category_data[:name])
```

db/seeds/002_categories.rb
```diff
@@ -0,0 +1,29 @@
...
+  { name: 'Bread' },
+  { name: 'Pasta' },
+  { name: 'Fruit' },
+  { name: 'Vegetable' },
+  { name: 'Seafood' },
+  { name: 'Poultry' },
+  { name: 'Beef' },
+  { name: 'Pork' },
+  { name: 'Sauce' },
+  { name: 'Condiment' }
```
nfoos: I suggest sorting these alphabetically for ease of maintenance.  You could also segment them into groups if that makes more sense.

```
# Group A
{ name: 'Foo' },
{ name: 'Bar' },

# Group B
{ name: 'Baz' },
{ name: 'Qux' },
```

db/migrate/20240124135559_categories.rb
```diff
@@ -0,0 +1,13 @@
+class Categories < ActiveRecord::Migration[7.0]
+  def change
+    create_table :categories do |t|
+      t.string :name, null:false, index: { unique:true }
+      t.timestamps
+    end
+
+    create_join_table :recipes, :categories do |t|
```
nfoos, line 8: nice, I didn't know that `create_join_table` was a thing. I like this better than using `create_table` here

db/migrate/20240124135559_categories.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.index :recipe_id
+      t.index :category_id
+    end
```
nfoos, line 11: One thing that I would add is a unique index on category_id + recipe_id.  This will prevent someone from adding the same category to a recipe more than once.

```suggestion
    end
    
    add_index :categories_recipes, [:category_id, :recipe_id], unique: true
```

app/models/category.rb
```diff
@@ -0,0 +1,5 @@
+class Category < ApplicationRecord
+  validates :name, presence: true
+
+  has_and_belongs_to_many :recipes, join_table: 'categories_recipes', foreign_key: true
```
nfoos: I don't think  you need to specify the `join_table` or `foreign_key` here. I think it's actually breaking things.
```suggestion
  has_and_belongs_to_many :recipes
```

app/models/recipe.rb
```diff
@@ -1,5 +1,7 @@
 class Recipe < ApplicationRecord
   validates :name, presence: true
 
+  has_and_belongs_to_many :categories, join_table: 'categories_recipes', foreign_key: true
```
nfoos: I don't think you need to specify the `join_table` or `foreign_key` here.  I think it's actually breaking things.
```suggestion
  has_and_belongs_to_many :categories
```

app/controllers/categories_controller.rb
```diff
@@ -0,0 +1,69 @@
+class CategoriesController < ApplicationController
+  before_action :set_category, only: %i[show edit update destroy]
+
+  # GET /categorys or /categorys.json
+  def index
+    @categories = Category.all
+    @category = Category.all.by_name
```
nfoos: Not sure what this is for, but it doesn't work
```suggestion
```

db/seeds/002_categories.rb
nfoos, line 1: It would actually be better for this to run before the recipes seeds file.  You could rename this to `001_categories.rb` and the other to `002_recipes.rb`

nfoos: We can also update the recipes seeds file to set the categories. For example, I would probably do something like:
```
recipes_data = [
  {
    name: "Spaghetti Bolognese",
    # ... other fields
    categories: %w[Pasta Main],
  },
  # ... other recipe data
]

recipes_data.each do |recipe_data|
  recipe = Recipe.find_or_initialize_by(name: recipe_data[:name])

  categories = recipe_data.delete(:categories) { [] }
  recipe.category_ids = Category.where(name: categories).pluck(:id)

  recipe.update(recipe_data)
end
```

app/controllers/categories_controller.rb
CodeSmitty: I added this for testing so see if I could access categories from the recipes form without making a model form. Forgot to remove it.

app/controllers/categories_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+        format.json { render json: @category.errors, status: :unprocessable_entity }
+      end
+    end
+  end
+
+  # DELETE /categorys/1 or /categorys/1.json
+  def destroy
+    @category.destroy
+    respond_to do |format|
+      format.html { redirect_to categorys_url, notice: 'category was successfully destroyed.' }
```
nfoos: Any mention of `categorys` in this file should be `categories`. Most are comments, but this will fail.
```suggestion
      format.html { redirect_to categories_url, notice: 'category was successfully destroyed.' }
```

app/controllers/categories_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+      format.json { head :no_content }
+    end
+  end
+
+  private
+
+  # Use callbacks to share common setup or constraints between actions.
+  def set_category
+    @category = category.find(params[:id])
```
nfoos: Not sure what's going on here
```suggestion
    @category = Category.find(params[:id])
```

app/controllers/categories_controller.rb
nfoos, line 1: The only needed method in this file is the `create` one if you intend on supporting creation of new categories from the recipe form. The others won't be necessary unless/until you want to add a section in the app just for managing the categories.

app/views/categories/index.html.erb
nfoos, line 1: Not really necessary unless you want to manage categories separately.

app/views/categories/_form.html.erb
nfoos, line 1: Not really necessary unless you want to manage categories separately.

app/views/recipes/_form.html.erb
```diff
@@ -48,6 +48,15 @@
...
 
+   <%# <input class='form-control' list='categories' id='category_name' name='recipe[name]' value='<%= @category %> 
+  <datalist id="categories">
+        <% Category.all.each do |c| %>
+            <option> <%= c.name %> </option>
+        <% end %>
+    </datalist>
+
+   <%= form.text_field 'categories[name]', list:"categories" %>
+
```
nfoos: This will give you a basic multiselect dropdown
https://api.rubyonrails.org/classes/ActionView/Helpers/FormOptionsHelper.html
```suggestion
   <%= form.collection_select :category_ids, Category.all, :id, :name, {}, { multiple: true } %>

```

Make sure to update recipe_params in the recipes controller to include `category_ids: []`

db/seeds/001_categories.rb
```diff
@@ -0,0 +1,46 @@
+# db/seeds.rb
+categories_data = [
+  #group A
```
nfoos: lol, you took my "group A" a little to literally.  I meant more like

```
categories = [
  # Course
  'Appetizer',
  'Dessert',
  'Main',

  # Cuisine
  'American',
  'Greek',
  'Italian',
  'Mexican',
]
```

db/seeds/002_recipes.rb
```diff
@@ -12,6 +12,7 @@
     notes: "You can add Parmesan cheese on top for extra flavor.",
     rating: 4.5,
     is_favorite: true,
+    categories: ["Main" "Pasta" "Fruit"]
```
nfoos:
```suggestion
    categories: %w[Main Pasta Fruit]
```

config/routes.rb
```diff
@@ -3,4 +3,5 @@
   # Define your application routes per the DSL in https://guides.rubyonrails.org/routing.html
 
   root 'recipes#index'
+  get 'categories', to: 'categories#create'
```
nfoos: This isn't being used, but it should be a `post `instead of a `get`
```suggestion
  post 'categories', to: 'categories#create'
```

app/models/category.rb
```diff
@@ -0,0 +1,6 @@
+class Category < ApplicationRecord
+  validates :name, presence: true
+
+  has_and_belongs_to_many :recipes
+  accepts_nested_attributes_for :recipes
```
nfoos: Don't need to accept nested attributes here
```suggestion
```

app/models/category.rb
```diff
@@ -0,0 +1,6 @@
...
+end
```
nfoos, line 11: This would be a useful helper method for getting categories by names.
```suggestion

  def self.from_names(names)
    names.map do |name|
      find_or_create_by(name: name)
    end
  end
end
```
  The `self` here means that it is a class method rather than an instance method. Meaning you call it like:

```
categories = Category.from_names(%w[Main Side Appetizer NewThing])
```

app/models/category_recipe.rb
```diff
@@ -0,0 +1,5 @@
+class CategoryRecipe < ApplicationRecord
+  validates :name, presence: true
+
+  has_and_belongs_to_many :recipes
```
nfoos: These are incorrect. This model does not have a name and does not have many or belong to many recipes.
```suggestion
```

app/models/recipe.rb
```diff
@@ -1,5 +1,13 @@
 class Recipe < ApplicationRecord
   validates :name, presence: true
+  has_and_belongs_to_many :categories
+  accepts_nested_attributes_for :categories
```
nfoos: This is generally fine, but it isn't actually being used at the moment.
```suggestion
```

db/seeds/002_recipes.rb
```diff
@@ -168,11 +181,18 @@
...
     is_favorite: false,
+    categories: %w[Salad Appetizer Side]
   },
 ]
 
 recipes_data.each do |recipe_data|
-  Recipe.find_or_initialize_by(name: recipe_data[:name]).update(recipe_data)
+  recipe = Recipe.find_or_initialize_by(name: recipe_data[:name])
+  categories = recipe_data.delete(:categories) { [] }
+  recipe.category_ids = Category.where(name: categories).pluck(:id)
```
nfoos, line 191: We might want to create the categories if they don't exist.  The suggested `from_names` method on the `Category` model is useful here.
```suggestion
  category_names = recipe_data.delete(:categories) { [] }
  recipe.categories = Category.from_names(category_names)
```

app/controllers/recipes_controller.rb
```diff
@@ -22,7 +22,7 @@ def create
     @recipe = Recipe.new(recipe_params)
 
     respond_to do |format|
-      if @recipe.save
+      if @recipe.save_categories
```
nfoos:
```suggestion
      if @recipe.save
```

app/models/recipe.rb
```diff
@@ -1,5 +1,13 @@
...
+
+  def save_categories
+    categories.map do |category|
+      Category.find_or_create_by(name: category.name.strip)
+    end
+  end
```
nfoos: I'm not sure if this is doing what you think it might be.  The `categories` it is accessing now would be the categories that already exist on the recipe instance, which could possibly work to persist new instances of Categories that haven't been saved to the database yet. But there was nothing initializing the categories. Also, this isn't saving the recipe itself, so it can't replace the save method.
```suggestion
```

app/controllers/recipes_controller.rb
```diff
@@ -22,7 +22,7 @@ def create
     @recipe = Recipe.new(recipe_params)
```
nfoos: I think this is how I might implement this.  First it takes the 'categories' param which is a string of comma separated category names and converts that to an array of category name strings.

Then it uses a method on the Category model for finding/creating categories from a list of names, returning an array of Category model instances.  It takes that new array and merges that into the recipe_params hash (essentially replacing the 'categories' param) which is used to initialize the Recipe before saving.

The reason why `merge` is used here is because `recipe_params` is a method call that rebuilds the params hash each time, so any changes to it would be lost.  `merge` is just taking two hashes and smashing them together to a new hash.
```suggestion
    category_names = recipe_params.fetch('categories', '').split(',').map(&:strip)

    @recipe = Recipe.new(recipe_params.merge( 'categories' => Category.from_names(category_names) }))
```

Similar logic could be used for the update route as well.  Although, you'd need to figure out how to populate your form element with the recipe's current categories when it loads so that the user doesn't always have to enter all the categories.

Further refactoring, I'd possibly update the Recipe model to handle category_names itself so that the seeds, controllers and other calling code don't need to be responsible for any of that.

app/views/recipes/_form.html.erb
```diff
@@ -48,6 +48,16 @@
...
   </div>
 
+    
+  <datalist id="categories">
+    <% Category.all.each do |c| %>
+        <option> <%= c.name %> </option>
+    <% end %>
+  </datalist>
+ 
+     <%= form.text_field ":categories", value: "", list: "categories", placeholder: "Enter categories, seperated by commas" %>
```
nfoos: The name can either be "categories" or :categories, but not ":categories"
```suggestion
     <%= form.text_field "categories", value: "", list: "categories", placeholder: "Enter categories, seperated by commas" %>  
```

CodeSmitty: This was me trying different combinations, because when I submit I would get errors. One in the console or 'unpermitted parameter' even though :categories is in the recipe params. Or I would get instance of {} or [] error.

app/controllers/recipes_controller.rb
CodeSmitty: This is a comment for both this and the model function of save categories. The reason I had @recipe.save_categories, (I did forget to add save in the model, but that was the intention). Going back into the model, then yea it definitely is not working like I imagine it would. I thought on submitting recipe it would go to the controller call `@recipe.save_categories` and in the model, it would map categories submitted by the form. Not categories already in the table and it would cross reference with the Category table and either find or create the category and save the recipe accordingly.

app/models/category.rb
```diff
@@ -0,0 +1,11 @@
+class Category < ApplicationRecord
+  validates :name, presence: true
+
+  has_and_belongs_to_many :recipes
+
+  def self.from_names(names)
```
nfoos, line 6: This needs a test. I will help with this.

app/controllers/categories_controller.rb
nfoos, line 1: Probably don't need this file

app/javascript/application.js
```diff
@@ -1,4 +1,4 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
 import "@hotwired/turbo-rails"
 import "controllers"
-import 'flowbite'
```
nfoos, line 4: I think this is needed for the navbar

app/models/recipe.rb
```diff
@@ -1,5 +1,15 @@
 class Recipe < ApplicationRecord
   validates :name, presence: true
+  has_and_belongs_to_many :categories
+  accepts_nested_attributes_for :categories
+
+  def category_names
```
nfoos, line 5: This needs a test. I will help with this.

app/models/recipe.rb
```diff
@@ -1,5 +1,15 @@
...
+    categories.pluck(:name)
+  end
+
+  def category_names=(category_names)
```
nfoos, line 9: This needs a test

app/views/categories/index.html.erb
nfoos, line 1: This file is not needed

app/views/layouts/application.html.erb
```diff
@@ -7,13 +7,14 @@
...
   </head>
+  <header>
+    <%= render "layouts/navigation"%>
+  </header>
   <body class="bg-gray-50">
-    <header>
-      <%= render "layouts/navigation"%>
-    </header>
-    <main class="container mx-auto mt-32 mb-16 px-5 flex">
+    <main class="container mx-auto mt-28 px-5 flex">
```
nfoos: Undo these changes. I assume this was a merge conflict with my latest branch.

app/views/recipes/_recipe.html.erb
```diff
@@ -24,6 +24,11 @@
     </div>
   </div>
 
+  <% recipe.categories.each do |c| %>
+    <h2><%= c.name %>,</h2>
+  <% end %>
+  
+
```
nfoos: This could be displayed better

db/seeds/002_recipes.rb
nfoos: Actually, this can be simplified now with category_names

app/views/icons/mini/_filled_heart.html.erb
nfoos, line 1: Is this being used?

app/models/category.rb
```diff
@@ -0,0 +1,11 @@
+class Category < ApplicationRecord
+  validates :name, presence: true
+
+  has_and_belongs_to_many :recipes
```
nfoos, line 4: You should pull in the rubocop config changes I made so it doesn't complain about this

config/routes.rb
nfoos: Should just remove it

config/importmap.rb
```diff
@@ -5,4 +5,4 @@
 pin "@hotwired/stimulus", to: "stimulus.min.js"
 pin "@hotwired/stimulus-loading", to: "stimulus-loading.js"
 pin_all_from "app/javascript/controllers", under: "controllers"
-pin "flowbite", to: "https://cdnjs.cloudflare.com/ajax/libs/flowbite/2.2.1/flowbite.turbo.min.js"
```
nfoos, line 8: I think this is needed for the navbar

app/models/category_recipe.rb
nfoos: Should just remove this entire file.

app/models/recipe.rb
nfoos: Still should remove this

db/seeds/002_recipes.rb
nfoos: Should fix this

app/views/recipes/_form.html.erb
```diff
@@ -42,10 +42,10 @@
       <% end %>
     </div>
   </div>
-  
-  <div class="my-5 flex flex-row justify-between w-1/6">
-    <%= form.label "#{:is_favorite}?" %>
-    <%= form.check_box :is_favorite, rows: 4, class: "block shadow rounded-md border border-gray-200 outline-none px-2 py-2 mt-1" %>
```
nfoos: I don't think you meant to remove this yet.

app/views/recipes/_recipe.html.erb
nfoos: https://flowbite.com/docs/components/badge/

db/seeds/001_categories.rb
nfoos: I'd say just remove the comments and extra lines for now

app/views/recipes/_form.html.erb
CodeSmitty: I got a little ahead of myself, was still thinking about new branch.

app/views/icons/mini/_filled_heart.html.erb
CodeSmitty: This file is for the is favorite click. It was not supposed to be in here yet.

app/views/recipes/show.html.erb
```diff
@@ -3,7 +3,11 @@
     <% if notice.present? %>
       <p class="py-2 px-3 bg-green-50 mb-5 text-green-500 font-medium rounded-lg inline-block" id="notice"><%= notice %></p>
     <% end %>
+
     <%= render @recipe %>
+    
+    <div class="flex flex-row flex-wrap">
+  </div>
```
nfoos: What is this for?

CodeSmitty: I missed a div lol. It was when I was testing the categories as a sidebar.

nfoos (approved):
**LGTM** 🛳️

PR #15, opened 2024-02-20
Recipe import

nfoos:
## Description

Provide a way to upload and download recipes in bulk. This is useful for copying from one environment to another and builds the foundation for restoring exports from other services. It can also serve as a personal backup.

image

Download button generates a ZIP archive of all of the logged in user's recipes
image

New upload page where user can upload an archive that was previously downloaded
image

After selecting a file and clicking upload, the file is processed then user is redirected to recipe index with a flash message
image

When saving a recipe, if one is already found with the same name for that user, it is skipped
image

Issues with processing the archive are shown via flash message as well
image


app/controllers/recipes_controller.rb
```diff
@@ -14,6 +18,33 @@ def web_result
...
+
+  # POST /recipes/archive/upload
+  def upload_archive
+    uploaded = params.require(:file)
+
+    result = Recipes::Archive.new(current_user).restore(uploaded.tempfile)
+
+    redirect_to recipes_url, notice: "Imported #{result[:created]} recipes (#{result[:skipped]} skipped)"
+  rescue StandardError => e
+    redirect_to recipes_url, alert: e.message
```
nfoos: Should probably limit display of error message to known Import errors and display generic message for any others

lib/recipes/archive/recipe_keeper.rb
```diff
@@ -0,0 +1,31 @@
...
+        'recipes.html'
+      end
+
+      def extract_recipes(contents)
+        document = Nokogiri::HTML(contents)
+        document.css('.recipe-details')
+      end
+
+      def import_recipe(recipe_doc, zip_file)
+        recipe = Recipes::Import::RecipeKeeper.new(recipe_doc).recipe
```
nfoos: load_recipe

CodeSmitty (approved):

PR #16, opened 2024-02-25
Import Recipes from Web Search

nfoos:
## Description

https://trello.com/c/8rGlFe5D/17-recipe-import-from-external-websites

Adds a way to import recipes from websites.

Uses Google's Programmable Search Engine to provide a search bar for finding recipes from specific websites. When a search result is clicked, instead of opening the website, the URL is passed to the backend to scrape the recipe info and fill out the recipe create form. The user can then choose to modify the recipe and/or save it.

When the recipe is saved, the URL is stored in the `source` field and a link icon is added to the recipe show page.

recipe_import


app/views/recipes/import.html.erb
```diff
@@ -0,0 +1,15 @@
+<script async src="https://cse.google.com/cse.js?cx=46a4f48e3fd6e4651"></script>
```
nfoos, line 1: Maybe make this configurable

CodeSmitty (dismissed):

CodeSmitty (dismissed):

nfoos:
@CodeSmitty Do you think this change should include a source field on the recipe to add the URL to the original recipe?

CodeSmitty:
Yes. I was thinking about that the other day. Was gonna mention that and the image being an option. The image could be a me thing though. For work related reasons, choosing a recipe based on how it looks and diffulty is a big deal for me. 

So having a source url is a lot help in the meantime. Thosr are my personal reasons. For UX, I think it will also help for the same reasons or just general seeing it from a diff perspective.

nfoos:
@CodeSmitty 👍 Yeah, generally the import won't be capturing all of the context from the recipe so it seems it would be good to have a way to easily open it on the source website.

The image will be something that we can eventually include on the import, but we need to build out the functionality of attaching images to recipes first.

CodeSmitty (approved):

PR #25, opened 2024-03-28
Filter recipe list

CodeSmitty:
Created a search bar for recipes.


app/controllers/recipes_controller.rb
```diff
@@ -14,7 +14,7 @@ def web_result
 
   # GET /recipes or /recipes.json
   def index
-    @recipes = Recipe.all.sorted
+    @recipes = Recipe.search(params)
```
nfoos:
```suggestion
    @recipes = Recipe.search(params).sorted
```

app/controllers/recipes_controller.rb
```diff
@@ -30,7 +30,7 @@ def edit; end
 
   # POST /recipes or /recipes.json
   def create
-    @recipe = Recipe.new(recipe_params)
+    @recipe = Recipe.new(find_recipe_params)
```
nfoos:
```suggestion
    @recipe = Recipe.new(recipe_params)
```

app/controllers/recipes_controller.rb
```diff
@@ -82,4 +82,8 @@ def recipe_params
     params.require(:recipe).permit(:name, :ingredients, :directions, :yield, :prep_time, :cook_time, :description,
                                    :rating, :is_favorite, :notes, :source, category_names: [])
   end
+
+  def find_recipe_params
+    params.require(:recipe).permit(:name, :ingredients, :directions, :notes, category_names: [])
+  end
```
nfoos:
```suggestion
```

app/javascript/controllers/filter_controller.js
```diff
@@ -0,0 +1,13 @@
+import { Controller } from "@hotwired/stimulus"
+import debounce from "lodash/debounce"
+
+export default class extends Controller {
+  initialize() {
+    this.submit = debounce(this.submit.bind(this), 1000)
+  }
+
+  submit(event) {
+    event.preventDefault()
```
nfoos: Don't think this is needed
```suggestion
```

app/javascript/controllers/filter_controller.js
```diff
@@ -0,0 +1,13 @@
+import { Controller } from "@hotwired/stimulus"
+import debounce from "lodash/debounce"
+
+export default class extends Controller {
+  initialize() {
+    this.submit = debounce(this.submit.bind(this), 1000)
```
nfoos: Can probably go with a lower value here
```suggestion
    this.submit = debounce(this.submit.bind(this), 300)
```

app/views/recipes/index.html.erb
```diff
@@ -2,9 +2,19 @@
...
-  <div id="recipes" class="flex-column flex-wrap justify-center lg:flex-nowrap">
-    <% @recipes.each do |recipe| %>
-      <%= link_to recipe.name, recipe, class: "block m-auto lg:w-1/3 font-medium mb-1 text-2xl p-4 text-center lg:text-left" %>
+  <div class="m-1">
+    <%= form_tag @recipe_path, method: :get, data: { turbo_frame: :recipes, turbo_action: "advance", controller: "filter", action: "input->filter#submit" } do |f| %>
+      <div class="flex flex-row flex-nowrap justify-center">
+        <%= text_field_tag "query", nil, placeholder: "Search ...", class: "block rounded-lg px-2.5 pb-2.5 w-1/2 text-sm text-gray-900 bg-gray-50  border border-teal-900 appearance-none focus:outline-none focus:ring-0 focus:border-teal-600 peer", autocomplete: 'off' %>
+        <%= submit_tag "Search", class: "inline-flex items-center rounded-lg py-1 px-2 bg-teal-900 hover:bg-white text-white hover:text-teal-900 border border-teal-900 inline-block font-medium cursor-pointer" %>
+      </div>
     <% end %>
```
nfoos: - use `recipes_path` instead of `@recipe_path`
- pass current query to input value
- autofocus on input element
- removed button, but can keep if you think it's necessary
- moved turbo-action from form to input

```suggestion
    <%= form_tag recipes_path, method: :get, data: { turbo_frame: :recipes, turbo_action: "advance", controller: "filter" } do |f| %>
      <div class="flex flex-row flex-nowrap justify-center">
        <%= text_field_tag :query, params[:query], data: { action: "input->filter#submit" }, placeholder: "Search ...", class: "block rounded-lg px-2.5 pb-2.5 w-1/2 text-sm text-gray-900 bg-gray-50  border border-teal-900 appearance-none focus:outline-none focus:ring-0 focus:border-teal-600 peer", autocomplete: 'off', autofocus: 'true' %>
      </div>
    <% end %>
```

app/views/recipes/index.html.erb
```diff
@@ -2,9 +2,19 @@
...
   </div>
+  <turbo_frame_tag :recipes data-turbo-if="@recipes.any?" />
+    <div id="recipes" class="flex-column flex-wrap justify-center lg:flex-nowrap">
+      <% @recipes.each do |recipe| %>
+        <%= link_to recipe.name, recipe, class: "block m-auto lg:w-1/3 font-medium mb-1 text-2xl p-4 text-center lg:text-left" %>
+      <% end %>
+    </div>
+  </turbo_frame_tag>
```
nfoos: Fixed turbo frame
```suggestion
  <%= turbo_frame_tag :recipes do %>
    <div id="recipes" class="flex-column flex-wrap justify-center lg:flex-nowrap">
      <% @recipes.each do |recipe| %>
        <%= link_to recipe.name, recipe, class: "block m-auto lg:w-1/3 font-medium mb-1 text-2xl p-4 text-center lg:text-left" %>
      <% end %>
    </div>
  <% end %>
```

nfoos (changes requested):

app/assets/stylesheets/application.css
```diff
@@ -57,3 +57,19 @@ textarea,
 .rating > input:not(:checked) ~ label:hover ~ label {
   background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 576 512'%3E%3C!--!Font Awesome Free 6.5.1 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2024 Fonticons, Inc.--%3E%3Cpath fill='%2363E6BE' d='M259.3 17.8L194 150.2 47.9 171.5c-26.2 3.8-36.7 36.1-17.7 54.6l105.7 103-25 145.5c-4.5 26.3 23.2 46 46.4 33.7L288 439.6l130.7 68.7c23.2 12.2 50.9-7.4 46.4-33.7l-25-145.5 105.7-103c19-18.5 8.5-50.8-17.7-54.6L382 150.2 316.7 17.8c-11.7-23.6-45.6-23.9-57.4 0z'/%3E%3C/svg%3E%0A");
 }
+
+input[type="search"]::-webkit-search-cancel-button {
```
nfoos: Is any of this needed?  Seems to work without it.

app/views/icons/mini/_magnifying.html.erb
```diff
@@ -0,0 +1,3 @@
+<svg class="w-4 h-4 text-gray-500 dark:text-gray-400" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 20 20">
```
nfoos: Pass in the custom text color through `classes`
```suggestion
<svg class="w-4 h-4 <%= classes %>" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 20 20">
```

CodeSmitty: Only reason it's set that way is because I copy pasted tailwind icon lol. I'll change it.

app/views/icons/mini/_magnifying.html.erb
nfoos, line 1: Should probably remove this file if it's not being used

app/assets/images/recipe.png
nfoos, line 1: I noticed the One_piece_logo.png file next to this one that should probably be deleted

app/views/recipes/index.html.erb
```diff
@@ -2,14 +2,36 @@
...
+              <div class="absolute inset-y-0 start-0 flex items-center ps-3 pointer-events-none">
+                <%= image_tag "recipe.png", class: "w-7/12" %>
+              </div>
+              <%= text_field_tag :query, params[:query], data: { filter_target: "query" }, placeholder: "Search ...", class: "w-full block rounded-lg px-2.5 pb-2.5 ps-10 text-sm text-gray-900 bg-gray-50 border border-teal-900 appearance-none focus:outline-none focus:ring-0 focus:border-teal-600 peer", autocomplete: 'off', autofocus: 'true', type: 'search', onfocus: "this.setSelectionRange(this.value.length,this.value.length);" %>
+          </div>
+          <div>
+            <%= select_tag(:category_names, options_from_collection_for_select(Category.all, :name, :name), { data: { controller: "tom-select", 'tom-select-create-value': false }, multiple: true, placeholder: "Select Categories" }) %>
+          </div>
+          <div class="query-btn flex justify-end py-4">
+            <%= submit_tag 'submit', class:"float-right inline-flex items-center rounded-lg py-1 px-2 bg-white hover:bg-teal-600 text-teal-900 hover:text-white border border-teal-900 inline-block font-medium cursor-pointer" %>
```
nfoos: Do you want to have a submit button? I know previously we were doing auto submit on input change.

app/models/recipe.rb
```diff
@@ -10,5 +10,24 @@ def category_names=(category_names)
     self.category_ids = Category.from_names(category_names).pluck(:id)
   end
 
+  scope :search, -> (params){
```
nfoos: Rubocop wants the method call rather than literal syntax for multi-line functions

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/Lambda

```
scope :search, lambda do |params|

end
```

**nitpick**
Also, I'm not a fan of the use of `params` as the argument name here as it might be misleading.  I think it can just be called `query`

app/models/recipe.rb
```diff
@@ -10,5 +10,24 @@ def category_names=(category_names)
...
+    return if params.blank?
+
+    query = "%#{sanitize_sql_like(params)}%"
+    where('recipes.name ILIKE ?', query)
+      .or(where('recipes.directions ILIKE ?', query))
+      .or(where('recipes.ingredients ILIKE ?', query))
+      .or(where('recipes.notes ILIKE ?', query))
+  }
+
+  scope :category_search, ->(category_names){
```
nfoos: **nitpick**

The naming of both of these scopes could be better. Often they are given adjective-like names rather than verbs, so they are describing the collection rather than being a list of actions.

Maybe something like:

`Recipe.with_text(query).in_categories(categories)`

app/javascript/controllers/filter_controller.js
nfoos, line 1: You'll want to run prettier on this file and the tom select controller.

`npx prettier --write .`

app/javascript/controllers/filter_controller.js
```diff
@@ -0,0 +1,20 @@
...
+    this.submit = debounce(this.submit.bind(this), 300);
+  }
+
+  connect() {
+    this.queryTarget.addEventListener("input", this.submit);
+    this.categoryTarget.addEventListener("input", this.submit);
+  }
+
+  submit() {
+    console.log('form=submit')
```
nfoos: debug log should be removed

app/javascript/controllers/filter_controller.js
```diff
@@ -0,0 +1,20 @@
+import { Controller } from "@hotwired/stimulus"
+import debounce from "lodash/debounce"
```
nfoos, line 2: This controller is failing to register because you removed the importmap for lodash. This also brings me back to the question of whether you want the form to auto-submit and couldn't get it working or if you would rather just use the button.

app/javascript/controllers/filter_controller.js
```diff
@@ -2,11 +2,19 @@ import { Controller } from "@hotwired/stimulus"
...
+  static targets = ["query", "category"];
+
   initialize() {
-    this.submit = debounce(this.submit.bind(this), 300)
+    this.submit = debounce(this.submit.bind(this), 300);
+  }
+
+  connect() {
+    this.queryTarget.addEventListener("input", this.submit);
+    this.categoryTarget.addEventListener("input", this.submit);
```
nfoos: Having the targets and using connect to add the event listeners like this is ok, but you can just add the listener from the HTML element like this and not have to manage the targets in the controller here.

`data: { action: "input->filter#submit" }`

nfoos (changes requested):

app/javascript/controllers/filter_controller.js
CodeSmitty: That is how I had it originally, and it wasn't working. (Before I added the lodash importmap). I added it now, and removed the connect. Still not working properly, but I am actually getting a trigger.

spec/models/recipes_spec.rb
```diff
@@ -21,35 +21,75 @@
...
+      let(:recipe) { described_class.create(name: 'Birria Tacos') }
+      let(:query) { 'taco' }
+
+      it { is_expected.to include recipe }
+      it { is_expected.to contain_exactly recipe }
+    end
+  end
+
+  describe '.in_categories' do
+    it 'returns all when param is empty'
```
nfoos: Just a reminder that these tests need to be finished

nfoos (approved):

PR #36, opened 2025-02-20
Search filter

CodeSmitty:


app/assets/images/recipe.png
nfoos, line 1: Looking at this made me notice this image which we should probably delete https://github.com/CodeSmitty/mame-zefu/blob/b6e4ad4091634b6f400c696fce1f10f0c01384c1/app/assets/images/One_piece_logo.png

app/assets/stylesheets/application.css
```diff
@@ -17,7 +17,7 @@
 input,
 textarea,
 .ts-control {
-  box-shadow: inset -1px -1px 2px 0 rgba(13, 148, 137, 0.94) !important;
+  box-shadow: 1px 2px 4px 1px rgba(206, 205, 205, 0.94) !important;
```
nfoos, line 20: I can't really tell what the difference is here 🤷

CodeSmitty: The shadow instead of going inwards into the input now it's just outward. I changed the colors from tea to simple gray and just changed the size of the shadow and blur.

PR #43, opened 2025-03-04
Import repairs

CodeSmitty:
I updated recipe imports for 

- All Recipes
- Taste of Home
- Food and Wine

## Description

I updated the class names for recipe details on all imports listed. For yield, prep time and cook time that seemed enough to fix the changes. 

I updated class names for Ingredients and Directions. 
Food and wine I removed entire old logic for ingredients and targeted P tag inside ul elements. 

## Testing

I looked up multiple recipes for each import. Every import seemed to work correctly on my end. The only one I did not get to fix was cook_time for Food and Wine.

### DIFF
--- a/lib/recipes/import/food_and_wine.rb
+++ b/lib/recipes/import/food_and_wine.rb
@@ -15,20 +15,42 @@ module Recipes
         recipe_details['total_time']
       end
 
-      def recipe_ingredients
+      def recipe_ingredients # rubocop:disable Metrics
         document
-          .css('ul.mm-recipes-structured-ingredients__list li p > span')
-          .map { |t| t.text.strip }
-          .compact_blank
-          .join("\n")
+          .css('div.mm-recipes-structured-ingredients')
+          .children
+          .reduce('') do |text, node|
+            next text unless node.element?
+
+            if node.name == 'ul'
+              text << node.children.map { |e| e.text.strip }.compact_blank.join("\n")
+            elsif node.name == 'p' && node.text.present?
+              text << "\n\n" if text.present?
+
+              text << "#{node.text.upcase}\n"
+            end
+
+            text
+          end
       end
 
-      def recipe_directions
+      def recipe_directions # rubocop:disable Metrics
         document
-          .css('li.mntl-sc-block-group--LI > p')
-          .map { |t| t.text.strip }
-          .compact_blank
-          .join("\n\n")
+          .css('div.mm-recipes-steps__content')
+          .children
+          .reduce('') do |text, node|
+            next text unless node.element?
+
+            if node.name == 'ol'
+              text << node.children.map { |e| e.css('> p').text.strip }.compact_blank.join("\n\n")
+            elsif node.name == 'h3' && node.text.present?
+              text << "\n\n" if text.present?
+
+              text << "#{node.text.strip.upcase}\n"
+            end
+
+            text
+          end
       end
 
       private
--- a/lib/recipes/import/food_and_wine.rb
+++ b/lib/recipes/import/food_and_wine.rb
@@ -15,42 +15,20 @@ module Recipes
         recipe_details['total_time']
       end
 
-      def recipe_ingredients # rubocop:disable Metrics
+      def recipe_ingredients
         document
-          .css('div.mntl-structured-ingredients')
-          .children
-          .reduce('') do |text, node|
-            next text unless node.element?
-
-            if node.name == 'ul'
-              text << node.children.map { |e| e.text.strip }.compact_blank.join("\n")
-            elsif node.name == 'p' && node.text.present?
-              text << "\n\n" if text.present?
-
-              text << "#{node.text.upcase}\n"
-            end
-
-            text
-          end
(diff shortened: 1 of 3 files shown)


lib/recipes/import/taste_of_home.rb
```diff
@@ -43,8 +51,8 @@ def recipe_directions
 
       def total_time
         document
-          .css('div.total-time p')
-          .text
+          .css('div.recipe-time span')
+          .map(&:text)
       end
     end
   end
```
nfoos: Doesn't look like this is being used anymore, so I'd just delete it.

**edit**: Actually I guess your code is using it, but my suggestion does not use it.

lib/recipes/import/taste_of_home.rb
```diff
@@ -3,26 +3,34 @@ class Import
...
+
+        total_time[cook_time + 1]
       end
 
       def recipe_total_time
-        total_time[%r{Prep/Total Time: (\d+ \w+)}, 1]
+        document
+          .css('div.time-unit')
+          .text[/(\d+ \w+)/, 1]
       end
```
nfoos: Looks like this is technically grabbing the cook time. We aren't actually storing this in the database. I think we could probably just remove this method.

lib/recipes/import/taste_of_home.rb
```diff
@@ -3,26 +3,34 @@ class Import
...
+
+        total_time[prep_time + 1]
       end
 
       def recipe_cook_time
-        total_time[/(?:Cook|Bake): (\d+ \w+)/, 1]
+        cook_time = total_time.index('Cook Time')
+        return nil unless cook_time
+
+        total_time[cook_time + 1]
```
nfoos: These appear to have unique class names, so I think they could be simplified a bit.

```
document
  .css('div.cook span')
  &.last
  &.text
```

lib/recipes/import/taste_of_home.rb
```diff
@@ -3,26 +3,34 @@ class Import
...
+          .css('div.yield span')
+          .text[/(\d+ \w+)/, 1]
       end
 
       def recipe_prep_time
-        total_time[/Prep: (\d+ \w+)/, 1]
+        prep_time = total_time.index('Prep Time')
+        return nil unless prep_time
+
+        total_time[prep_time + 1]
```
nfoos: These appear to have unique class names, so I think they could be simplified a bit.
```
document
  .css('div.prep span')
  &.last
  &.text
```

lib/recipes/import/taste_of_home.rb
```diff
@@ -3,26 +3,34 @@ class Import
...
+          .css('h2.recipe-title')
           .text
       end
 
       def recipe_yield
         document
-          .css('div.makes p')
-          .text
+          .css('div.yield span')
+          .text[/(\d+ \w+)/, 1]
```
nfoos: Following the same approach as my other suggestions
```
document
  .css('div.yield span')
  &.last
  &.text
```

lib/recipes/import/food_and_wine.rb
```diff
@@ -15,54 +15,32 @@ def recipe_total_time
         recipe_details['total_time']
       end
 
-      def recipe_ingredients # rubocop:disable Metrics
+      def recipe_ingredients
         document
-          .css('div.mntl-structured-ingredients')
```
nfoos, line 20: I think you only needed to change the class name here to `div.mm-recipes-structured-ingredients`. The way it was implemented before was meant to capture multiple lists with headings

For example:
image

The new changes aren't really handling the ingredients very well at all

image

Some recipes I used to to test
- https://www.foodandwine.com/recipes/rib-eye-aguachile
- https://www.foodandwine.com/recipes/spice-roasted-pork-tenderloin
- https://www.foodandwine.com/lemony-shrimp-dumplings-6422735

lib/recipes/import/food_and_wine.rb
```diff
@@ -15,54 +15,32 @@ def recipe_total_time
...
+          .css('ul.mm-recipes-structured-ingredients__list li p > span')
+          .map { |t| t.text.strip }
+          .compact_blank
+          .join("\n")
       end
 
-      def recipe_directions # rubocop:disable Metrics
+      def recipe_directions
         document
-          .css('div.recipe__steps-content')
```
nfoos, line 39: I think you could have just changed this class here to `div.mm-recipes-steps__content` and then, I'll comment below the change needed to exclude the figcaptions

lib/recipes/import/food_and_wine.rb
```diff
@@ -15,54 +15,32 @@ def recipe_total_time
...
-          .children
-          .reduce('') do |text, node|
-            next text unless node.element?
-
-            if node.name == 'ol'
-              text << node.children.map { |e| e.text.strip }.compact_blank.join("\n\n")
```
nfoos, line 45: This should work:
`text << node.children.map { |e| e.css('> p').text.strip }.compact_blank.join("\n\n")`

nfoos (approved):

PR #48, opened 2025-03-28
navigation template 1

CodeSmitty:
First template of Navigation bar. 
## Testing

This is mostly a desktop version. Mobile is not tested well, but it works on mobile for time being. Will update mobile version after desktop is more or less done.


app/views/icons/mini/_sign_in.html.erb
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="<%= classes %>" class="size-6 <%= classes %>">
```
nfoos: What values are you adding to the `classes` variable that is being used by `stroke`.  If there are distinct values that we want to pass to `class` and `stroke` I think we'd want to use a new variable.

CodeSmitty: I removed stroke from the image. I had trouble getting stroke to work with tailwind classes why I added the classes. We could use a different variable for stroke. I don't think it's needed. Since stroke, stroke width still work on the icon with the classes. Up to you if you would rather give it, it's own variable.

nfoos (approved):

PR #49, opened 2025-04-02
User authorization

CodeSmitty:
Added User authorization with pundit for user recipes.


Gemfile
```diff
@@ -89,3 +89,5 @@ group :test do
   gem "capybara"
   gem "selenium-webdriver"
 end
+
+gem "pundit", "~> 2.5"
```
nfoos: Should probably move this up with the other gems not in groups

app/controllers/recipes_controller.rb
```diff
@@ -15,23 +15,23 @@ def web_result
 
   # GET /recipes or /recipes.json
   def index
-    @recipes = Recipe.with_text(params[:query]).in_categories(params[:category_names]).sorted
+    @recipes = policy_scope(Recipe.with_text(params[:query]).in_categories(params[:category_names]).sorted)
```
nfoos: Should be able to chain this like policy_scope(Recipe).with_text...

app/controllers/recipes_controller.rb
```diff
@@ -15,23 +15,23 @@ def web_result
...
   end
 
   # GET /recipes/1 or /recipes/1.json
   def show; end
 
   # GET /recipes/new
   def new
-    @recipe = Recipe.new
+    @recipe = current_user.recipes.build
```
nfoos: I think you can use new instead of build. They do the same thing, I'm just used to new

app/models/user.rb
```diff
@@ -1,3 +1,4 @@
 class User < ApplicationRecord
   include Clearance::User
+  has_many :recipes, dependent: nil
```
nfoos: Why the dependent: nil?

app/policies/application_policy.rb
nfoos, line 1: Were the contents of this file auto generated?

app/policies/recipe_policy.rb
nfoos, line 1: I think you'll need show and index methods as well

app/views/recipes/index.html.erb
```diff
@@ -21,6 +21,7 @@
     <% end %>
   </div>
   <div id="recipes" class="flex-column flex-wrap justify-center lg:flex-nowrap">
+    <% if @recipes %>
```
nfoos: This shouldn't be necessary if @recipes is an array. Is it nil if there are no recipes for the user?

db/migrate/20250330181923_add_user_ref_to_recipes.rb
```diff
@@ -0,0 +1,6 @@
+class AddUserRefToRecipes < ActiveRecord::Migration[7.2]
+  Recipe.delete_all
```
nfoos: Is there a reason why this isn't inside the method? Does it work like this?

db/seeds/002_recipes.rb
```diff
@@ -195,6 +209,7 @@
     description: "Kebabs are a quintessential summer recipe, and our version got a sweet upgrade with fresh pineapple, red bell pepper, and a mixture of honey, soy sauce, and apple cider vinegar. If you have a bit of extra time, you could marinate the kebabs first, but it's equally delicious brushed on while grilling.",
     notes: "If using wooden skewers, soak them in water for 30 minutes before threading to avoid them charring or catching on fire.\n\nTry using steak instead of  pork and mango instead of pineapple.",
     category_names: %w[Pork Main],
+    user_id: 1
```
nfoos: Not a fan of using hardcoded ids. You should find the existing user instead. Either User.first or find by email. Also, since all recipes are getting the same user, you can just set it on the loop below

user.recipes.find_or_initialize...

Gemfile
CodeSmitty: I did not look at it. My mistake. I just installed it didn't pay attention to it.

app/models/user.rb
CodeSmitty: Rubocop

app/policies/application_policy.rb
CodeSmitty: yes, I needed to run rails g pundit:install. I wasn't sure if I wanted to create it myself. So I generated it, less chance of mistake on my part. Reading docs, I noticed a lot of it ran from application_policy. Why I added it. (It worked without it too).

app/views/recipes/index.html.erb
CodeSmitty: Yes, it is nil if there are no recipes. I added that so I could see the application. The code broke when I added the scope because I hadn't implemented the scope correctly exeg(scope.where(user=user). So I got cannot read each of nil error. I know if @recipes is not needed. It's temporary.

db/migrate/20250330181923_add_user_ref_to_recipes.rb
CodeSmitty: The reason is, I'm dumb and wasn't sure where to put it; but yes it works like this. lol

spec/factories/clearance.rb
```diff
@@ -4,7 +4,18 @@
   end
 
   factory :user do
-    email
+    email { generate(:email) }
```
nfoos: This change is not necessary. The email sequence works without it
image

spec/factories/clearance.rb
```diff
@@ -4,7 +4,18 @@
...
+FactoryBot.define do
+  sequence :name do |n|
+    "chicken#{n}"
+  end
+
+  factory :recipe do
+    name { :name }
+    user { :user }
+  end
+end
```
nfoos: I'd move this to it's own file. It technically has nothing to do with `clearance` and I usually see each model with its own factory file.  So this would go in **spec/factories/recipe.rb**.  Honestly, you could probably rename this file as well to be **spec/factories/user.rb**

spec/factories/clearance.rb
```diff
@@ -4,7 +4,18 @@
...
 end
+
+FactoryBot.define do
+  sequence :name do |n|
+    "chicken#{n}"
+  end
+
+  factory :recipe do
+    name { :name }
+    user { :user }
```
nfoos: What you have here won't work.

``` suggestion
    name
    user
```

spec/factories/clearance.rb
```diff
@@ -4,7 +4,18 @@
...
   factory :user do
-    email
+    email { generate(:email) }
     password { 'password' }
   end
 end
+
+FactoryBot.define do
+  sequence :name do |n|
+    "chicken#{n}"
```
nfoos: If you want to have some default realistic values, you could install and use the faker gem https://github.com/faker-ruby/faker

spec/policies/recipe_policy_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+require './spec/support/features/clearance_helpers'
+
+RSpec.describe RecipePolicy, type: :request do
+  let(:user) { create(:user) }
+  let(:other_user) { create(:user) }
+  let(:recipe) { create(:recipe, user: user) }
+
+  context 'when unauthenticated' do
+    it 'redirects to login' do
+      get '/recipes'
```
nfoos: The tests you've added to this file don't belong here. They should be in the recipe request spec. This file would be used to test the policy class directly without any http request

spec/factories/clearance.rb
CodeSmitty: I actually read a lot about faker. Wasn't sure if you wanted to go that route.

app/policies/recipe_policy.rb
```diff
@@ -0,0 +1,53 @@
+class RecipePolicy < ApplicationPolicy
+  attr_reader :user, :recipe
+
+  def initialize(user, recipe)
+    @user = user
+    @recipe = recipe
+  end
+
+  def show
```
nfoos: Got the wrong name, which means the method in `ApplicationPolicy ` is being used instead
```suggestion
  def show?
```

app/policies/recipe_policy.rb
```diff
@@ -0,0 +1,53 @@
...
+    owner?
+  end
+
+  def index
```
nfoos: Wrong name here as well. I'm also skeptical of the `owner?` check for this action since you will have a collection of recipes instead of just one. Technically, you aren't even calling authorize on the index.  I'd probably just make this return `true`.
```suggestion
  def index?
```

spec/requests/recipes_spec.rb
```diff
@@ -1,25 +1,70 @@
...
+    context 'when unauthenticated' do
+      it 'redirects to login' do
+        get '/recipes'
+        expect(response).to redirect_to(sign_in_path)
+      end
     end
   end
 
   describe 'GET /recipes/:id' do
-    it 'returns a 200' do
```
nfoos, line 19: Not sure why you deleted this test

spec/requests/recipes_spec.rb
```diff
@@ -1,25 +1,70 @@
...
   let(:recipe_name) { 'Spaghetti' }
   let(:category_name) { 'Italian' }
   let(:category) { Category.create(name: category_name) }
-  let!(:recipe) { Recipe.create(name: recipe_name, categories: [category]) }
-  let(:user) { create(:user) }
+  let!(:recipe) { Recipe.create(name: recipe_name, categories: [category], user: user) }
+  let(:other_user) { create(:user) }
 
   describe 'GET /recipes' do
-    it 'returns a 200' do
```
nfoos, line 11: Not sure why you deleted this test

app/policies/recipe_policy.rb
CodeSmitty: This is one of those that confuse me lol. When is should I use the ? or not. I read it in the docs but still have trouble deciding.

app/controllers/application_controller.rb
```diff
@@ -4,6 +4,9 @@
 
 class ApplicationController < ActionController::Base
   include Clearance::Controller
+  include Pundit::Authorization
+  after_action :verify_authorized unless :health_check
```
nfoos: This isn't working how you think it might be.  The symbol `:health_check` evaluates to true, so it will never apply the `after_action`.

The proper syntax would be `after_action :verify_authorized, except: %i[health_check]`

However, you'll find that the `index` route will raise an exception since `authorize` is not being called.  The pundit README provides a good example to enforce `policy_scope` on `index` and `authorize` on all other actions. Combined with an exception for `health_check` it would look like this:

```
after_action :verify_pundit_authorization, except: %i[health_check new create]

...

private

def verify_pundit_authorization
  if action_name == "index"
    verify_policy_scoped
  else
    verify_authorized
  end
end
```

app/controllers/recipes_controller.rb
```diff
@@ -15,23 +15,26 @@ def web_result
...
   def index
-    @recipes = Recipe.with_text(params[:query]).in_categories(params[:category_names]).sorted
+    @recipes = policy_scope(Recipe).with_text(params[:query]).in_categories(params[:category_names]).sorted
   end
 
   # GET /recipes/1 or /recipes/1.json
-  def show; end
+  def show
+    authorize @recipe
+    @recipe = Recipe.find(params[:id])
```
nfoos: This isn't needed, `@recipe` was already set by the `set_recipe` before action.  It also wouldn't make sense to set the recipe after calling authorize, because you could end up setting a different recipe.

I think you could handle all of the `authorize` calls in this controller by updating the `set_recipe` method:

```
def set_recipe
  @recipe = authorize Recipe.find(params[:id])
end
```

spec/requests/recipes_spec.rb
```diff
@@ -1,25 +1,103 @@
...
+        expect { get recipe_path(recipe) }.to raise_error(Pundit::NotAuthorizedError)
+      end
+    end
+  end
+
+  describe 'POST /recipes' do
+    context 'when user is unauthenticated' do
+      it 'redirects to login' do
+        post '/recipes'
+        expect { post recipes_path }.to redirect_to(sign_in_path)
```
nfoos:
```suggestion
        expect(response).to redirect_to(sign_in_path)
```

spec/factories/recipe.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  sequence :name do |n|
+    "chicken#{n}"
+  end
+
+  factory :recipe do
+    name
+    association :user
```
nfoos: I had to look this up. The `association` here is optional, it works implicitly since the factory and association names are the same https://github.com/thoughtbot/factory_bot/blob/main/GETTING_STARTED.md#implicit-definition

db/seeds/003_users.rb
nfoos, line 1: The seeds files are ran in alphanumeric order. You'll want the users seed to run before the other two, so you should rename them accordingly.

001_users.rb
002_categories.rb
003_recipes.rb

db/seeds/003_users.rb
```diff
@@ -0,0 +1,8 @@
+# db/seeds.rb
+exit unless Rails.env.development?
+
+admin = User.find_or_initialize_by(email: 'admin@example.com')
+admin.password = 'password'
+admin.save!(validate: false)
```
nfoos: I think you can condense this down to one line:
```suggestion
User.find_or_initialize_by(email: 'admin@example.com').update(password: 'password')
```

db/migrate/20250330181923_add_user_ref_to_recipes.rb
nfoos: Interesting. So I think the way this is, it will delete all the recipes when rolling back as well as migrating.  If you wanted to only delete the recipes on migrate, you'd want to do something like:

```
def up
  Recipe.delete_all
  add_reference :recipes, :user, null: false, foreign_key: true
end

def down
  remove_reference :recipes, :user, null: false, foreign_key: true
end
```

app/models/user.rb
nfoos: I think I'd probably set this to `:destroy`, although I'd also be open to `:restrict_with_error` or `:restrict_with_exception`

app/controllers/recipes_controller.rb
```diff
@@ -46,6 +46,7 @@ def create
 
   # PATCH/PUT /recipes/1 or /recipes/1.json
   def update
+    redirect_to :edit, notice: 'This user is not allowed to update this recipe.' unless authorize @recipe
```
nfoos: This redirect doesn't actually happen because `authorize` raises an error. You can also remove this `authorize` call since `set_recipe` already does it.

You might consider handling the errors as described here https://github.com/varvet/pundit?tab=readme-ov-file#rescuing-a-denied-authorization-in-rails

app/controllers/recipes_controller.rb
```diff
@@ -67,6 +69,7 @@ def destroy
   end
 
   def toggle_favorite
+    authorize @recipe
```
nfoos: Already handled by `set_recipe`

app/models/user.rb
CodeSmitty: I think I prefer destroy, maybe expeption?? If user gets deleted, I would assume we want the recipes to with with them.

db/migrate/20250330181923_add_user_ref_to_recipes.rb
CodeSmitty: So do I need to do a rollback on this specific migration and re migrate to fix this?

nfoos: That would be the proper way to do it, yes.

You could also checkout the main branch, reset the db, then switch back to this branch, update the migration script and then apply

app/policies/recipe_policy.rb
nfoos: Common practice is to name methods that return true/false with a ?

Here specifically, I think pundit is inferring that if you are calling authorize from an action named foo, it should use the foo? method in the policy class

app/controllers/recipes_controller.rb
```diff
@@ -59,6 +59,7 @@ def update
 
   # DELETE /recipes/1 or /recipes/1.json
   def destroy
+    redirect_to :show, notice: 'This user is not allowed to delete this recipe.' unless authorize @recipe
```
nfoos: I guess I should have commented on this line as well. You can remove it because authorize is already being called and the redirect won't happen because it raises an error

CodeSmitty: I just missed it. I looked around for any other authorize. My bad.

app/controllers/application_controller.rb
```diff
@@ -4,6 +4,9 @@
 
 class ApplicationController < ActionController::Base
   include Clearance::Controller
+  include Pundit::Authorization
+  after_action :verify_pundit_authorization, except: %i[health_check new create]
```
nfoos: After thinking about this for a while, it would probably be best to not exclude `new` and `create` here. Rubocop doesn't like it anyway.

Removing it from here, you will need to skip authorization for in the recipe and user controllers.

app/controllers/recipes_controller.rb
nfoos, line 1: You need to skip authorization for `web_search` and `web_result`

`skip_after_action :verify_pundit_authorization, only: %i[web_search web_result new create]`

app/policies/recipe_policy.rb
```diff
@@ -0,0 +1,53 @@
...
+    owner?
+  end
+
+  def index?
+    true
+  end
+
+  def create?
+    owner?
+  end
```
nfoos: I'd maybe just remove these 2 methods since they aren't actually being used

app/policies/recipe_policy.rb
```diff
@@ -0,0 +1,53 @@
+class RecipePolicy < ApplicationPolicy
+  attr_reader :user, :recipe
+
+  def initialize(user, recipe)
+    @user = user
+    @recipe = recipe
+  end
```
nfoos: You can just remove these, and then update the `owner?` method to use `record` instead of `recipe`. You might even want to consider moving the `owner?` method to the `ApplicationPolicy` because it would be useful for other policies in the future.

spec/models/recipes_spec.rb
```diff
@@ -1,12 +1,14 @@
 require 'rails_helper'
+require './spec/support/features/clearance_helpers'
```
nfoos: I don't think this is needed

spec/models/recipes_spec.rb
```diff
@@ -1,12 +1,14 @@
...
 
 RSpec.describe Recipe do
   it { is_expected.to validate_presence_of(:name) }
 
   describe '#category_names' do
     subject { recipe.category_names }
 
-    let(:recipe) { described_class.create(name: 'Spaghetti') }
+    let(:user) { create(:user) }
+    let(:recipe) { described_class.create(name: 'Spaghetti', user: user) }
```
nfoos: I think I'd rather just use factory bot here and not worry about the user at all.
```suggestion
    let(:recipe) { create(:recipe, name: 'Spaghetti') }
```

spec/models/recipes_spec.rb
```diff
@@ -23,7 +25,9 @@
   describe '#category_names=' do
     subject(:described_method) { recipe.category_names = category_names }
 
-    let(:recipe) { described_class.create(name: 'Spaghetti') }
+    let(:user) { create(:user) }
+
+    let(:recipe) { described_class.create(name: 'Spaghetti', user: user) }
```
nfoos:
```suggestion
    let(:recipe) { create(:recipe, name: 'Spaghetti') }
```

spec/models/recipes_spec.rb
```diff
@@ -61,9 +65,10 @@
   describe '.with_text' do
     subject { described_class.with_text(query) }
 
+    let(:user) { create(:user) }
```
nfoos:
```suggestion
```

app/controllers/application_controller.rb
CodeSmitty: So would do I just disable the line then? Leave them there.

spec/models/recipes_spec.rb
```diff
@@ -61,9 +65,10 @@
...
     let!(:recipes) do
```
nfoos: Factory bot has a nice way of creating multiple records
`let!(:recipes) { create_list(:recipe, 3) }`

spec/models/recipes_spec.rb
```diff
@@ -80,7 +85,7 @@
     end
 
     context 'when query does not match case' do
-      let(:recipe) { described_class.create(name: 'Birria Tacos') }
+      let(:recipe) { described_class.create(name: 'Birria Tacos', user: user) }
```
nfoos:
```suggestion
      let(:recipe) { create(:recipe, name: 'Birria Tacos') }
```

spec/models/recipes_spec.rb
```diff
@@ -91,14 +96,15 @@
   describe '.in_categories' do
     subject { described_class.in_categories(category_names) }
 
+    let(:user) { create(:user) }
```
nfoos:
```suggestion
```

spec/models/recipes_spec.rb
```diff
@@ -91,14 +96,15 @@
...
     let!(:japanese) { Category.create!(name: 'Japanese') }
 
-    let!(:tacos) { described_class.create!(name: 'Tacos', categories: [mexican]) }
-    let!(:pasta) { described_class.create!(name: 'Pasta', categories: [italian]) }
-    let!(:sushi) { described_class.create!(name: 'Sushi', categories: [japanese]) }
-    let!(:fusion) { described_class.create!(name: 'Fusion Dish', categories: [mexican, japanese]) }
+    let!(:tacos) { described_class.create!(name: 'Tacos', categories: [mexican], user: user) }
+    let!(:pasta) { described_class.create!(name: 'Pasta', categories: [italian], user: user) }
+    let!(:sushi) { described_class.create!(name: 'Sushi', categories: [japanese], user: user) }
+    let!(:fusion) { described_class.create!(name: 'Fusion Dish', categories: [mexican, japanese], user: user) }
```
nfoos:
```suggestion
    let!(:tacos) { create(:recipe, name: 'Tacos', categories: [mexican]) }
    let!(:pasta) { create(:recipe, name: 'Pasta', categories: [italian]) }
    let!(:sushi) { create(:recipe, name: 'Sushi', categories: [japanese]) }
    let!(:fusion) { create(:recipe, name: 'Fusion Dish', categories: [mexican, japanese]) }
```

app/controllers/application_controller.rb
CodeSmitty: Never mind, I missunderstood what you meant.

spec/requests/recipes_spec.rb
```diff
@@ -1,25 +1,168 @@
...
-      expect(response).to have_http_status(:ok)
+    context 'when authenticated' do
+      it 'returns a 200' do
+        get recipe_path(recipe, as: user)
+        expect(response).to have_http_status(:ok)
+      end
+    end
+
+    context 'when athenticated and recipe non owner' do
+      before { sign_in(other_user) }
```
nfoos: Rather than actually signing in, clearance provides a backdoor middleware that lets you specify the user with the `as` parameter. https://github.com/thoughtbot/clearance?tab=readme-ov-file#fast-feature-specs

I'm not sure which I prefer. This can save on the total test runtime because you are reducing the number of requests. Whichever you prefer, it would be best to be consistent across all tests.

spec/requests/recipes_spec.rb
```diff
@@ -1,25 +1,168 @@
 require 'rails_helper'
+require './spec/support/features/clearance_helpers'
```
nfoos: I don't think this is needed

spec/requests/recipes_spec.rb
```diff
@@ -1,25 +1,168 @@
...
+
+        expect { delete recipe_path(recipe) }.to raise_error(Pundit::NotAuthorizedError)
+      end
+    end
+  end
+end
+
+  private
+
+def sign_in(user)
```
nfoos: This method could be moved to a helper module in spec/support/requests

nfoos (approved):

PR #50, opened 2025-04-14
Upgraded Tailwind css to v4

CodeSmitty:
## Testing

I tested the category inputs with tom select. Everything else had the correct styling. Those were the only broken ones. I added application.tailwind.css file to stylesheets. Fixed the issue, I moved it it to the newly created tailwind folder and it broke. So that will need to live there. I checked all pages for any styling changes. All seemed unchanged. 

I removed classes from the form inputs and added back to see the changes were happening.


app/assets/tailwind/application.css
```diff
@@ -0,0 +1,75 @@
+@import 'tailwindcss';
+@import 'https://cdnjs.cloudflare.com/ajax/libs/tom-select/2.3.1/css/tom-select.min.css'
+  layer(utilities);
```
nfoos: It bugs me a bit that `app/assets/stylesheets/application.tailwind.css` still exists, but even more so that the styling is duplicated between the two files. As best I can tell, it must be an issue with this import not actually working.

app/assets/tailwind/application.css
```diff
@@ -0,0 +1,75 @@
...
+@plugin '@tailwindcss/aspect-ratio';
+@plugin '@tailwindcss/typography';
+
+@theme {
+  --color-primary: rgb(18 18 18);
+  --color-secondary: #eeb14a;
+  --color-tertiary: rgb(165 68 29);
+  --color-input-label: rgb(238 177 74 / 73%);
+  --color-nav-backdrop: rgb(45 45 45);
+  --color-nav-link-bg: rgba(0, 0, 0, 0.4);
```
nfoos, line 13: The duplication of these from the tailwind.config.js is also not great

app/assets/tailwind/application.css
CodeSmitty: I can test by removing the import. I manually created this file because all of the styling for tom-select and I believe web import. I couldn't get the styling to work with using classes on the form. Unless you want to use javascript instead??? Not sure how best to accomplish this.

app/assets/tailwind/application.css
CodeSmitty: I could remove the --color and name them how we had them if that works better.

nfoos: I'm not really concerned about the naming, I just see these same colors defined in 2 separate files. Are they both being used? If we want to update one, do both files need to be modified?

app/assets/tailwind/application.css
nfoos: It would be nice if the styling would work with just setting classes on the form, but really my problem is just that the same CSS is defined in 2 places. Surely, it's not actually needed in both places, right?  I think I was able to remove everything from `app/assets/stylesheets/application.tailwind.css` except for the tom-select import and then remove the tom-select import from this file and everything appeared to work the same.

app/views/layouts/_navigation.html.erb
nfoos, line 1: Did the hover effect stop working on the navbar links?

app/assets/tailwind/application.css
CodeSmitty: So for this I removed the stylings from both files and nothing changed. The styles were still there. With both of them remoevd. So I am very confused here.

app/views/layouts/_navigation.html.erb
CodeSmitty: The hover still works on my end.

app/assets/tailwind/application.css
CodeSmitty: Something ```tailwind/application.css``` I removed import everything works. Restarted server the next line ```layer(utilities);``` Doesn't let the server start. Get a invalid declaration: layer etc. If I removed that all together. I get cannot apply class: text secondary. Soooo yea...

CodeSmitty: Update: Removing import and layer utilities from ```tailwind/application.css``` and removing the styles from the ```tailwind.config.js``` Seems to fix the issue. So now we would only need to add/change color themes in the ```tailwind/application.css```

app/javascript/controllers/tom_select_controller.js
```diff
@@ -27,7 +27,7 @@ export default class extends Controller {
       })
     document
       .getElementById("recipe_category_names-ts-control")
```
nfoos: **non-blocking**:

I'm seeing errors logged to the console for `addEventListener` being called on `null`.  I'm not sure what might have changed, but it looks like the element you are trying to select has the id of `category_names-ts-control` instead of `recipe_category_names-ts-control`.

I know this isn't exactly related to the tailwind upgrade, but it's probably worth fixing now.

nfoos (dismissed):
Sorry for letting this get so stale

nfoos (dismissed):

nfoos (approved):

PR #57, opened 2025-12-30
Use lograge for production logging

nfoos:
## Description

Lograge provides a way to filter out the health_check route from logs, which is nice because that route is used pretty extensively by kubernetes and uptime monitors.

This also helps consolidate the request logs to fewer lines.


nfoos:
I think semantic logger does a better job

PR #59, opened 2025-12-31
fixed broken recipe imports

CodeSmitty:
Fixed broken recipe imports.


app/views/recipes/_form.html.erb
```diff
@@ -16,6 +16,15 @@
     <%= form.label :name, class: "absolute text-sm text-gray-500 duration-300 transform -translate-y-4 scale-75 top-2 z-10 origin-[0] bg-gray-50 px-2 peer-focus:bg-gray-50 peer-focus:pb-0 peer-focus:px-2 peer-focus:text-tertiary peer-placeholder-shown:scale-100 peer-placeholder-shown:-translate-y-1/2 peer-placeholder-shown:top-6 peer-focus:top-2 peer-focus:scale-75 peer-focus:-translate-y-4 peer-focus:rtl:translate-x-1/4 peer-focus:rtl:left-auto start-1" %>
   </div>
 
+  <div class="my-5 lg:my-2 basis-full sm:basis-auto relative">
```
nfoos: Did you mean to move this? Do you think the image should be moved for the show page as well?

CodeSmitty: Yes, I can move the show as well. I liked the image under the tittle. Similar to how most websites do. Yield, prep, etc under the image. At least for now.

nfoos (approved):

PR #60, opened 2026-01-03
integraded image import for remaining recipe pages

CodeSmitty:
Added image imports to all recipes pages.

### DIFF
--- a/lib/recipes/import/all_recipes.rb
+++ b/lib/recipes/import/all_recipes.rb
@@ -9,6 +9,13 @@ module Recipes
           .text
       end
 
+      def recipe_image_url
+        document
+          .css('div.photo-dialog__page div#photo-dialog__item_1-0 img')
+          .map { |e| e.attr('src').freeze }
+          .join
+      end
+
       def recipe_yield
         "Serves #{recipe_details['servings']}"
       end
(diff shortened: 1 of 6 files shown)


lib/recipes/import/all_recipes.rb
```diff
@@ -9,6 +9,13 @@ def recipe_name
           .text
       end
 
+      def recipe_image_url
+        document
+          .css('div.photo-dialog__page div#photo-dialog__item_1-0 img')
+          .map { |e| e.attr('src').freeze }
+          .join
```
nfoos: I prefer the first/attr pattern from the taste of home class over this map/join pattern, which only works because there is only one matching element. If there were a second, it would be combining all of the urls which we saw the other day

The same comment applies to the other recipe classes

lib/recipes/import/food_network.rb
```diff
@@ -8,6 +8,14 @@ def recipe_name
           .strip
       end
 
+      def recipe_image_url
+        document
+          .css('section.o-RecipeLead img')
+          .map { |e| e.attr('src').freeze }
+          .join
+          .gsub(%r{\A//}, 'https://')
```
nfoos: This regex is essentially doing the same thing as the insert from the gordon ramsay class. I'd like for them to be consistent. The insert may be simpler than the regex

lib/recipes/import/all_recipes.rb
CodeSmitty: I changed all other to be more consistent. I did not change this one, because of the way I am pulling the image. Using first was not working. I believe this way is safer for this specific instance because of the way all recipes has their images laid out.

nfoos: I think this would work. That targets the primary image at the top of the page. It looks like what you have was targeting the collection of photos in the footer that show when you click "Photos" near the recipe name.

```
        document
          .css('figure.primary-image img')
          .first
          &.attr('src')
```

nfoos (approved):

PR #62, opened 2026-01-09
Update search results links

nfoos:
## Description

Updates the stimulus controller which overrides the click action for search result links. This change was prompted to support image search results.

All links now go to the web result import page. A link to the recipe source was added to the import page.

image

https://github.com/user-attachments/assets/4ca6dccb-0268-468d-8740-e06339d93729

### DIFF
--- a/app/javascript/controllers/search_results_controller.js
+++ b/app/javascript/controllers/search_results_controller.js
@@ -2,34 +2,70 @@ import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
   connect() {
+    this._abort = new AbortController()
+
+    if (!this._onClickLink) this._onClickLink = this.onClickLink.bind(this)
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/search_results_controller.js
```diff
@@ -21,14 +21,39 @@ export default class extends Controller {
...
+              const href = this.getAttribute("href")
+              if (href) {
+                e.preventDefault()
+
+                const params = new URLSearchParams({
+                  url: this.getAttribute("href"),
+                })
+
+                window.location.href = `web_result?${params}`
+              }
```
nfoos: I don't like how this code is duplicated, but I couldn't figure out how to extract it into a re-usable method

CodeSmitty (approved):

PR #68, opened 2026-01-11
Image edit upload

CodeSmitty:
Adding functionality for user to upload image from device into a recipe. 
Adding functionality for user to replace image from an existing recipe.


config/routes.rb
```diff
@@ -24,6 +24,7 @@
   resources :recipes do
     member do
       post :toggle_favorite
+      delete :delete_image
```
nfoos: I think I'd prefer the route to be `DELETE /recipes/:id/image` instead of `DELETE /recipes/:id/delete_image`.

The controller action can name can still remain the same

```suggestion
      delete :image, action: :delete_image
```
You'll have to update the path helper used by the button in the form

.rubocop.yml
```diff
@@ -24,6 +24,9 @@ Metrics/BlockLength:
   Exclude:
     - spec/**/*
 
+Metrics/AbcSize:
+  Max: 22
+
```
nfoos: I'd prefer disabling the cop inline instead of increasing the limit.

app/controllers/recipes_controller.rb
```diff
@@ -109,6 +111,14 @@ def toggle_favorite
     render json: { is_favorite: @recipe.is_favorite }
   end
 
+  def delete_image
+    @recipe = Recipe.find(params[:id])
```
nfoos: This isn't needed. The `set_recipe `before action is already doing this.

```suggestion
```

app/controllers/recipes_controller.rb
```diff
@@ -1,10 +1,10 @@
 class RecipesController < ApplicationController
   before_action :require_login
-  before_action :set_recipe, only: %i[show edit update destroy toggle_favorite]
+  before_action :set_recipe, only: %i[show edit update destroy toggle_favorite delete_image]
   skip_after_action :verify_pundit_authorization, only: %i[
     web_search web_result
     download_archive upload_archive_form upload_archive
-    new create
+    new create delete_image
```
nfoos: This should definitely not be added. The `set_recipe` before action is calling `authorize` so it's not needed. And it would be very concerning if we did not call authorize to verify the user owns the recipe for the image being deleted

```suggestion
    new create
```

app/models/recipe.rb
```diff
@@ -40,6 +44,21 @@ def category_names=(category_names)
 
   private
 
+  def log_destruction
+    Rails.logger.error "=== RECIPE #{id} IS BEING DESTROYED! ==="
+    Rails.logger.error caller.join("\n")
+  end
```
nfoos: Do you want to keep this? It seems like some debug logging, not sure what it gives us in prod

app/models/recipe.rb
nfoos, line 1: Looks like all of the changes to this file can be discarded when removing the `remove_image` checkbox/param

app/controllers/recipes_controller.rb
```diff
@@ -82,7 +82,10 @@ def create
   end
 
   # PATCH/PUT /recipes/1 or /recipes/1.json
-  def update
+  def update # rubocop:disable Metrics/MethodLength
+    uploaded = recipe_params[:image]
+    @recipe.image.purge if uploaded.present? && @recipe.image.attached?
```
nfoos: Pretty sure this isn't needed

app/javascript/controllers/delete_image_controller.js
```diff
@@ -0,0 +1,43 @@
...
+        method: "DELETE",
+        headers: {
+          "Content-Type": "application/json",
+          "X-CSRF-Token": token,
+          accept: "text/html, application/json",
+        },
+      })
+
+      console.log("Response status:", response.status)
+      console.log("Response headers:", response.headers)
```
nfoos: Don't think this logging is needed

app/controllers/recipes_controller.rb
```diff
@@ -119,6 +129,7 @@ def set_recipe
   # Only allow a list of trusted parameters through.
   def recipe_params
     params.require(:recipe).permit(:name, :ingredients, :directions, :yield, :prep_time, :cook_time, :description,
-                                   :rating, :is_favorite, :notes, :source, :image_src, category_names: [])
+                                   :rating, :is_favorite, :notes, :source, :image, :image_src,
+                                   :remove_image, :delete_image, category_names: [])
```
nfoos: `:remove_image` isn't needed anymore. Not sure what `:delete_image` was for but doesn't seem like it's being used

```suggestion
                                   category_names: [])
```

app/javascript/controllers/upload_image_controller.js
CodeSmitty, line 1: For when you look at this. I assume this is going to be one of the ones you want to simplify. 

What this does, is it creates a new input based on the old input and replaces it. If user deletes the selected image on replace image button click. 

The delete button is incase user decides not to replace image, can save form with changes to other inputs without needing to refresh page. 

The issue

- If user deletes replacement image and then selects a different one. When user submits form, it  actually saves the last selected image. Not the current image. This fixes that.  

Encountered issues just 'clearing input' because of browser and rails security best practices not allowing it.

app/views/icons/mini/_attachment.erb
```diff
@@ -0,0 +1,9 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 64 64" aria-labelledby="title"
```
nfoos: Where did you get this icon? I've been exclusively using https://heroicons.com/

spec/requests/recipes_spec.rb
```diff
@@ -135,6 +155,36 @@
...
+        end.to change(ActiveStorage::Attachment, :count).by(-1)
+      end
+
+      it 'calls delete image action' do
+        recipe.image.attach(uploaded_image)
+
+        expect do
+          delete image_recipe_path(recipe, as: user)
+        end.to change(ActiveStorage::Attachment, :count).by(-1)
+      end
```
nfoos, line 189: This test doesn't belong within this describe block since it is for a different endpoint. It is also a duplicate of the one below.

```suggestion
```

spec/requests/recipes_spec.rb
```diff
@@ -135,6 +155,36 @@
...
+    end
+  end
+
+  describe 'DELETE #delete_image' do
```
nfoos: Rename for consistency

```suggestion
  describe 'DELETE /recipes/:id/image' do
```

spec/requests/recipes_spec.rb
```diff
@@ -135,6 +155,36 @@
...
+    let(:recipe) { create(:recipe, user: user) }
+
+    before do
+      recipe.image.attach(uploaded_image)
+    end
+
```
nfoos, line 195: We should also include an unauthenticated test and one for trying to delete the image for a recipe owned by another user

spec/requests/recipes_spec.rb
```diff
@@ -8,6 +8,15 @@
...
   let(:other_recipe) { create(:recipe, user: other_user) }
+  let(:image_path) { Rails.root.join('spec/fixtures/files/test.png') }
+  let(:uploaded_image) do
+    fixture_file_upload(
+      Rails.root.join('spec/fixtures/files/test.png'),
+      'image/png'
+    )
+  end
+
+  let(:recipe_with_image) { attributes_for(:recipe, image: uploaded_image) }
```
nfoos: You're not using `image_path` or `recipe_with_image` and `uploaded_image` does not require an absolute path.

```suggestion
  let(:uploaded_image) { fixture_file_upload('test.png', 'image/png') }
```

spec/requests/recipes_spec.rb
```diff
@@ -135,6 +155,36 @@
...
+      end
+    end
+  end
+
+  describe 'DELETE #delete_image' do
+    let(:recipe) { create(:recipe, user: user) }
+
+    before do
+      recipe.image.attach(uploaded_image)
+    end
```
nfoos: The recipe can be created with the recipe attached

```suggestion
    let(:recipe) { create(:recipe, user: user, image: uploaded_image) }
```

spec/requests/recipes_spec.rb
```diff
@@ -62,6 +71,17 @@
...
+
+    it 'creates recipe then attaches image' do # rubocop:disable RSpec/ExampleLength
+      expect { post recipes_path(params: { recipe: attributes_for(:recipe) }, as: user) }
+        .to change(Recipe, :count).by(1)
+      recipe = Recipe.last
+      expect do
+        recipe.image.attach(io: Rails.root.join('spec/fixtures/files/test.png').open,
+                            filename: 'test.png',
+                            content_type: 'image/png')
+      end.to change(ActiveStorage::Attachment, :count).by(1)
```
nfoos: This isn't really testing image attachment in the context of an HTTP request. We _should_ be able to just pass the uploaded_image in the params, but I think you were running into the same issue that I am (and cannot find a solution that works)

```
     Failure/Error: @recipe = current_user.recipes.new(recipe_params)
     
     ActiveSupport::MessageVerifier::InvalidSignature:
       mismatched digest
```

I think I'd rather just remove this test since it pretty much just duplicates an existing test and then makes a false assertion.

spec/requests/recipes_spec.rb
CodeSmitty: I actually had it that way and kept getting mismatched digest errors. That's how I ended with that, I suspect it was conflicting with something I removed on the last round of conversions. I'll change it an test it again.

spec/requests/recipes_spec.rb
nfoos: We can also leave a placeholder test as a reminder to try fixing it later

```
it 'attaches the image', skip: 'ActiveSupport::MessageVerifier::InvalidSignature: mismatched digest' do
  expect { post recipes_path(params: { recipe: { name: recipe_name, image: uploaded_image } }, as: user) }
    .to change(ActiveStorage::Attachment, :count).by(1)
end
```

app/views/icons/mini/_attachment.erb
CodeSmitty: I honestly don't remember tbh. I changed the svg to match heroicons.

app/views/icons/mini/_attachment.erb
```diff
@@ -0,0 +1,9 @@
...
+aria-describedby="desc" role="img" xmlns:xlink="http://www.w3.org/1999/xlink" width="20px" height="20px">
+  <title>Attachment</title>
+  <desc>A line styled icon from Orion Icon Library.</desc>
+  <path data-name="layer1"
+  d="M54.5 29.3A16 16 0 0 0 31.9 6.7L8.3 30.2a2 2 0 0 0 2.9 2.8L34.7 9.5a12 12 0 0 1 17 17L22.5 55.7a8 8 0 1 1-11.3-11.3L37.5 18a4 4 0 1 1 5.7 5.7L22.5 44.4a2 2 0 0 0 2.8 2.8L46 26.5a8 8 0 1 0-11.3-11.3L8.3 41.5a12 12 0 0 0 17 17z"
+  fill="none" stroke="#eeb14a" stroke-miterlimit="10" stroke-width="2" stroke-linejoin="miter"
```
nfoos: https://github.com/user-attachments/assets/9200abac-0c11-4c0f-b4cc-6f21384f2136

CodeSmitty: Meant to comment not unresolve. Easy fix.

nfoos:
I think you already noticed this, but the styling isn't great on desktop

https://github.com/user-attachments/assets/0aa81c7e-6f9e-49e5-93c8-f6063cdf5251

app/assets/images/camera.png
nfoos, line 1: I'd be open to a different default image. I'm not a huge fan of the one I picked

app/views/recipes/_form.html.erb
```diff
@@ -16,14 +16,44 @@
...
-  </div>
+  <div
+    class="mx-auto my-5 lg:my-2 basis-full relative w-full lg:w-1/2"
+    data-controller="image"
+    data-image-delete-url-value="<%= delete_image_url(recipe) %>"
+    data-image-preview-src-value="<%= preview_image_src(recipe) %>"
+    data-image-persisted-src-value="<%= persisted_image_src(recipe) %>"
+    data-image-default-src-value="<%= default_image_src %>">
+    <div class="relative flex justify-end w-full md:w-10/12 lg:w-10/12 m-auto aspect-square overflow-hidden">
+      <button
```
nfoos: We probably don't want to show this button when the placeholder image is shown

app/views/recipes/_form.html.erb
```diff
@@ -16,14 +16,44 @@
...
+        type="button"
+        class="absolute rounded-lg py-2 px-2 bg-transparent hover:bg-secondary text-secondary hover:text-white font-medium cursor-pointer"
+        data-image-target="deleteButton"
+        data-action="image#deleteButtonClick:prevent">
+        <span><%= render "icons/micro/trash", classes: "me-1" %></span>
```
nfoos: Not necessary, but it might be nice if the icon were different between deleting a persisted image vs removing the uploaded, but not yet saved image (more like a cancel?) 🤷

app/helpers/recipes_helper.rb
```diff
@@ -1,2 +1,25 @@
...
+
+    image_recipe_path(recipe)
+  end
+
+  def preview_image_src(recipe)
+    if recipe.image_src.present?
+      recipe.image_src
+    elsif recipe.image.attached? && !recipe.image.persisted?
+      url_for(recipe.image)
+    end
```
nfoos: I was reviewing your tests and realized that the logic here is probably reversed. But then I realized that this helper is only used on page load, so it's not really possible for there to be an image attached that isn't persisted.

Then I realized these helpers aren't all that helpful lol

spec/helpers/recipe_helper_spec.rb
nfoos, line 1: These tests aren't bad, but I'm probably going to suggest discarding them, removing the  helper and just putting the logic inline in the form. I'll leave some comments here regardless for learning purposes

spec/helpers/recipe_helper_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'rails_helper'
+
+RSpec.describe 'RecipesHelper' do
+  let(:user) { create(:user) }
+  let(:uploaded_image) { fixture_file_upload('test.png', 'image/png') }
+  let!(:recipe) { create(:recipe, user: user, image: uploaded_image) }
```
nfoos: Slight refactor to this setup to support suggested changes
 - The `!` shouldn't be necessary since the tests that need the recipe are referencing that variable which will trigger it to be initialized
 -  Wait to attach the image to the recipe

```
  let(:image) { fixture_file_upload('test.png', 'image/png') }
  let(:recipe) { create(:recipe, user: user) }
```

spec/helpers/recipe_helper_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  describe '#delete_image_url' do
+    context 'when recipe image is persisted' do
+      before do
+        image = instance_double(ActiveStorage::Attachment, persisted?: true)
+        allow(recipe).to receive(:image).and_return(image)
+        allow(helper).to receive(:image_recipe_path).with(recipe).and_return('/recipes/1/image')
+      end
```
nfoos: I think this object mocking is unnecessary here. The test suite can handle attaching files and there's no external dependency. This mocking also has the risk of masking real errors 

```suggestion
      before do
        recipe.image = image
        recipe.save!
      end
```

spec/helpers/recipe_helper_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+      it 'returns the image deletion path' do
+        expect(helper.delete_image_url(recipe)).to eq('/recipes/1/image')
```
nfoos: With the change to actually saving the attachment, you can't rely on the recipe id being consistent, but you can reference the actual id

```suggestion
        expect(helper.delete_image_url(recipe)).to eq("/recipes/#{recipe.id}/image")
```

spec/helpers/recipe_helper_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      end
+    end
+
+    context 'when recipe image is not persisted' do
+      before do
+        image = instance_double(ActiveStorage::Attachment, persisted?: false)
+        allow(recipe).to receive(:image).and_return(image)
```
nfoos: This follows the same logic as when the image is persisted, without the save. This should reflect what is happening in the recipe controller when a user uploads a file but has not clicked save yet.

I think this is a lot easier to understand what is happening when reading through the tests

```suggestion
        recipe.image = image
```

spec/helpers/recipe_helper_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      end
+    end
+  end
+
+  describe '#preview_image_src' do
+    context 'when recipe has image_src' do
+      it 'returns the image source' do
+        expect(helper.preview_image_src(recipe)).to eq(recipe.image_src)
+      end
+    end
```
nfoos: This is actually not testing what is expected. image_src was never set, so it's actually comparing the return value when neither parts of the if statement are true

```
  def preview_image_src(recipe)
    if recipe.image_src.present?
      recipe.image_src
    elsif recipe.image.attached? && !recipe.image.persisted?
      url_for(recipe.image)
    end
  end
```

app/views/recipes/_form.html.erb
```diff
@@ -16,15 +16,45 @@
...
-      <%= image_tag @recipe.image_src, class: "preview-image" %>
-      <%= form.hidden_field :image_src, value: @recipe.image_src %>
-    <% end %>
+  <div
+    class="mx-auto my-5 lg:my-2 basis-full relative w-full lg:w-1/2"
+    data-controller="image"
+    data-image-delete-url-value="<%= delete_image_url(recipe) %>"
+    data-image-preview-src-value="<%= preview_image_src(recipe) %>"
+    data-image-persisted-src-value="<%= persisted_image_src(recipe) %>"
+    data-image-default-src-value="<%= default_image_src %>">
```
nfoos: Probably better off getting rid of the helpers after realizing the only complexity it was actually reducing is the check if the image is attached so the url_for method doesn't throw an error.

```suggestion
    data-image-delete-url-value="<%= image_recipe_path(recipe) %>"
    data-image-preview-src-value="<%= recipe.image_src %>"
    data-image-persisted-src-value="<%= url_for(recipe.image) if recipe.image.attached? %>"
    data-image-default-src-value="<%= asset_path('camera.png') %>">
```

app/views/recipes/_form.html.erb
CodeSmitty: Correct, I still haven't had time to touch styling. Wanted to make sire the tests were running.

app/views/recipes/_form.html.erb
CodeSmitty: I guess that is basically the intention for an attached image. It's a cancelation not a  deletion. Do you think the code is unnecessary, or should I change the text to say cancel instead? On image replacement.

app/helpers/recipes_helper.rb
CodeSmitty: Is that why I had so much trouble getting the image attach to work? I kept gettinf ActiveRecord::Attached, doesn't have method persisted or something like that. How I ended up with current code.

app/assets/images/camera.png
CodeSmitty: I personally prefer not be a default image lol, but yea I can look around.

app/helpers/recipes_helper.rb
nfoos: not sure 🤷

app/views/recipes/_form.html.erb
nfoos: I was thinking more along the lines of toggling the icon between the trash can and something else.

app/assets/images/camera.png
nfoos: If you want it without the placeholder you can tear that out. We can always add it back in later if we change our minds

spec/requests/recipes_spec.rb
```diff
@@ -174,6 +175,37 @@
...
         expect { delete recipe_path(recipe, as: other_user) }.to raise_error(Pundit::NotAuthorizedError)
       end
+
+      it 'deletes attached image when recipe is deleted' do
+        recipe.image.attach(uploaded_image)
+
+        expect do
+          delete recipe_path(recipe, as: user)
+        end.to change(ActiveStorage::Attachment, :count).by(-1)
+      end
```
nfoos:
```suggestion
```

app/views/recipes/_form.html.erb
```diff
@@ -16,16 +16,48 @@
...
-    <% if @recipe.image.attached? %>
-      <%= image_tag @recipe.image, class: "preview-image" %>
-    <% elsif @recipe.image_src.present? %>
-      <%= image_tag @recipe.image_src, class: "preview-image" %>
-      <%= form.hidden_field :image_src, value: @recipe.image_src %>
-    <% end %>
+  <div
+    class="mx-auto my-5 lg:my-2 basis-full relative w-full lg:w-1/2"
+    data-controller="image"
+    data-image-delete-url-value="<%= image_recipe_path(recipe) %>"
```
nfoos: This raises an error on the new recipe form because the recipe is not persisted and does not have an id, so the path cannot be successfully constructed

Something like this should work:
```suggestion
    data-image-delete-url-value="<%= image_recipe_path(recipe) if recipe.image.persisted? %>"
```

app/policies/recipe_policy.rb
```diff
@@ -15,6 +15,10 @@ def toggle_favorite?
     owner?
   end
 
+  def delete_image?
```
nfoos, line 18: There's now a recipe policy spec that you should update to check this permission

app/views/recipes/_form.html.erb
```diff
@@ -1,5 +1,5 @@
 
-<%= form_with(model: recipe, class: "contents") do |form| %>
+<%= form_with(model: recipe, class: "contents", local: true, data: { controller: "upload-image" }) do |form| %>
```
nfoos: The controller was renamed and attached below, so this change can be reverted. The data controller anyway, not sure about `local: true`

app/javascript/controllers/delete_image_controller.js
nfoos, line 1: The logic from this controller was added to the image controller, so this file can be deleted

app/views/recipes/_form.html.erb
```diff
@@ -16,16 +16,48 @@
...
-    <% elsif @recipe.image_src.present? %>
-      <%= image_tag @recipe.image_src, class: "preview-image" %>
-      <%= form.hidden_field :image_src, value: @recipe.image_src %>
-    <% end %>
+  <div
+    class="mx-auto my-5 lg:my-2 basis-full relative w-full lg:w-1/2"
+    data-controller="image"
+    data-image-delete-url-value="<%= image_recipe_path(recipe) %>"
+    data-image-preview-src-value="<%= recipe.image_src %>"
+    data-image-persisted-src-value="<%= url_for(recipe.image) if recipe.image.attached? %>"
```
nfoos: I ran into an edge case where this errored. I tried saving a recipe with an image but no name, so the page loaded with a validation error and I think technically the image might have been considered attached but not persisted so it wouldn't have a url?

```suggestion
    data-image-persisted-src-value="<%= url_for(recipe.image) if recipe.image.persisted? %>"
```

app/views/recipes/_form.html.erb
```diff
@@ -16,16 +16,48 @@
...
+    data-image-default-src-value="<%= asset_path('placeholder.jpg') %>"
+    data-image-state-value="<%= recipe.image.attached? ? 'persisted' : 'none' %>">
```
nfoos: I think I saw some weirdness here too. It looks like recipe.image.persisted? works best for all of these checks.

```suggestion
    data-image-state-value="<%= recipe.image.persisted? ? 'persisted' : 'none' %>">
```

app/views/recipes/_form.html.erb
```diff
@@ -16,16 +16,48 @@
...
+            <span id="delete-button-icon" class="hidden"><%= render "icons/micro/trash", classes: "me-1" %></span>
+            <span id="cancel-button-icon" class="hidden"><%= render "icons/outline/cancel", classes: "size-4" %></span>
+        </button>
+      <%= image_tag "", data: { 'image-target': 'previewImage' }, class: "w-full object-contain" %>
+    </div>
+    <div class="my-5 relative flex justify-center">
+      <button type="button"
+        id="upload-image-button"
+        class="inline-flex items-center rounded-lg py-2 px-3 bg-white hover:bg-secondary text-secondary hover:text-white border border-secondary inline-block font-medium cursor-pointer"
+        onclick="document.getElementById('upload-image-input').click();">
```
nfoos: I'd like to keep all the Javascript inside the stimulus controller
```suggestion
        data-action="image#uploadButtonClick:prevent">
```

app/javascript/controllers/image_controller.js
```diff
@@ -0,0 +1,102 @@
...
+      this.stateValue = this.persistedSrcValue ? "persisted" : "none"
+      this.updateUi()
+      console.log("state:value", this.stateValue)
+    } else if (this.persistedSrcValue) {
+      this.deletePersistedImage()
+      document.getElementById("upload-image-button").textContent =
+        "Upload Image"
+    }
+  }
+
```
nfoos, line 71:
```suggestion

  uploadButtonClick() {
    this.fileFieldTarget.click()
  }

```

app/javascript/controllers/image_controller.js
```diff
@@ -0,0 +1,102 @@
...
+      this.previewSrcValue = undefined
+      const newInput = this.fileFieldTarget.cloneNode(false)
+      this.fileFieldTarget.replaceWith(newInput)
+      this.stateValue = this.persistedSrcValue ? "persisted" : "none"
+      this.updateUi()
+      console.log("state:value", this.stateValue)
+    } else if (this.persistedSrcValue) {
+      this.deletePersistedImage()
+      document.getElementById("upload-image-button").textContent =
+        "Upload Image"
```
nfoos: I prefer to use targets when possible

```suggestion
      this.uploadButtonTarget.textContent = "Upload Image"
```

app/javascript/controllers/image_controller.js
```diff
@@ -0,0 +1,102 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["previewImage", "deleteButton", "fileField"]
```
nfoos:
```suggestion
  static targets = ["previewImage", "deleteButton", "uploadButton", "fileField"]
```

app/views/recipes/_form.html.erb
```diff
@@ -16,16 +16,48 @@
...
+          data-action="image#deleteButtonClick:prevent">
+            <span id="delete-button-icon" class="hidden"><%= render "icons/micro/trash", classes: "me-1" %></span>
+            <span id="cancel-button-icon" class="hidden"><%= render "icons/outline/cancel", classes: "size-4" %></span>
+        </button>
+      <%= image_tag "", data: { 'image-target': 'previewImage' }, class: "w-full object-contain" %>
+    </div>
+    <div class="my-5 relative flex justify-center">
+      <button type="button"
+        id="upload-image-button"
+        class="inline-flex items-center rounded-lg py-2 px-3 bg-white hover:bg-secondary text-secondary hover:text-white border border-secondary inline-block font-medium cursor-pointer"
```
nfoos:
```suggestion
        class="inline-flex items-center rounded-lg py-2 px-3 bg-white hover:bg-secondary text-secondary hover:text-white border border-secondary inline-block font-medium cursor-pointer"
        data-image-target="uploadButton"
```

app/javascript/controllers/image_controller.js
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+
+  deleteButtonClick() {
+    if (this.previewSrcValue) {
+      this.previewSrcValue = undefined
+      const newInput = this.fileFieldTarget.cloneNode(false)
+      this.fileFieldTarget.replaceWith(newInput)
+      this.stateValue = this.persistedSrcValue ? "persisted" : "none"
+      this.updateUi()
```
nfoos: This is redundant because of the change callback defined up above
```suggestion
```

nfoos:
I noticed an issue after saving recipes where the directions were no longer separated. Apparently, the existence of the file input field changes how data is encoded and submitted. Those line breaks are now getting saved as "\r\n\r\n" instead of "\n\n" which messes with the regex on the recipe show page.

The following can be added to the recipe model to address the issue
```
class Recipe < ApplicationRecord
  before_save :normalize_line_endings
  
  private
  
  def normalize_line_endings
    %w[directions ingredients notes].each do |field|
      text = send(field)
      self[field] = text&.gsub("\r\n", "\n")&.gsub("\r", "\n") if text.present?
    end
  end
end
```

app/views/recipes/_form.html.erb
nfoos: I think this one got missed

nfoos (approved):

PR #71, opened 2026-01-19
Increase code coverage

nfoos:
## Description

Increase total code coverage + various improvements for issues discovered while adding tests

## Testing

Added all kinds of tests


app/controllers/recipes_controller.rb
```diff
@@ -76,8 +76,8 @@ def create
         format.html { redirect_to recipe_url(@recipe), notice: 'Recipe was successfully created.' }
         format.json { render :show, status: :created, location: @recipe }
       else
-        format.html { render :new, status: :unprocessable_entity }
```
nfoos, line 79: I saw a message during testing that :unprocessable_entity was being deprecated

CodeSmitty (approved):

app/controllers/application_controller.rb
```diff
@@ -3,17 +3,24 @@
 require 'net/http'
 
 class ApplicationController < ActionController::Base
+  rescue_from ActionController::RoutingError, with: :not_found
+  rescue_from ActiveRecord::RecordNotFound, with: :not_found
```
nfoos, line 7: The not_found changes help with reducing noise in the logs which were getting full stack traces

app/models/recipe.rb
```diff
@@ -49,8 +49,7 @@ def attach_image_from_url
       filename: File.basename(image_src)
     )
   rescue Down::Error => e
-    errors.add(:image, "couldn't be downloaded: #{e.message}")
-    throw :abort
+    Rails.logger.error("Failed to download image from #{image_src}: #{e.message}")
```
nfoos, line 52: I figured it would be better to just log an error and still let the recipe save if there is an issue downloading the image

lib/recipes/import/gordon_ramsay.rb
```diff
@@ -30,9 +30,9 @@ def recipe_ingredients # rubocop:disable Metrics
             next text unless node.element?
 
             if node.name == 'ul'
-              text << node.children.map(&:text).join
+              text << node.children.select(&:element?).map(&:text).join("\n")
```
nfoos, line 33: This reduces some extra whitespace that could have been getting captured. I'm not sure it really affected production, but was annoying in the tests

lib/recipes/import/json_schema.rb
```diff
@@ -12,8 +12,18 @@ def recipe_name
...
+        return if recipe_json['image'].blank?
+
+        image_uri =
+          recipe_json['image']
+          .then { Array.wrap(_1) }
+          .first
+          .then { |image| image.is_a?(Hash) ? image['url'] : image }
+          .then { |url| URI(url) }
+
+        return unless image_uri.is_a?(URI::HTTP) || image_uri.is_a?(URI::HTTPS)
```
nfoos, line 24: Ignore when image url doesn't have protocol / domain name. I see this mostly happen when it is a default image.

lib/recipes/import/json_schema.rb
```diff
@@ -37,6 +47,8 @@ def recipe_category_names
       end
 
       def recipe_description
+        return if recipe_json['description'].blank?
```
nfoos, line 50: Return early so unescape doesn't fail if there is no description

PR #86, opened 2026-02-07
Yield calculator

CodeSmitty:
https://github.com/user-attachments/assets/de124290-d3e6-4e61-b426-4f6f80556992

Created a Calculator for Servings. 
User inputs number of servings
Yield Updater runs the call to the controller for IngredientCalculator. 

- Parser parses the ingredients texts and pulls the ingredient quantity and unit
- Scaler Seperates quantity and multiplies by user inserted multiplier. Then joins the result with the descripion of the ingredients. 
- Unit converter, converts the unit to the next unit threshold. Example 4 Cups of Ingredient = 1 qt of Ingredient
- The converter should convert up and down. Depending on user's use case. 

### Edge Cases
There are still a couple of bugs. Reason for merge is to live test and find any other edge cases that come up. 
I included one in the video. 

- There seems to be an issue where based on the saved conversion. The next conversion bugs on certains units. 
- pint conversion did not play well when scaling down. Removed for now, might add with future testing. 
- Video bug, conversions goes from 2 pound to 1 xxxxx/xxxxxx kg. First time seeing that error. Could be a one off, will need to test further. 
- Scaling back down _after_ recipe save is not perfect. During the same instance of recipe save. Most recipes scale up and down perfectly. On dom refresh, the new saved recipe yield and ingredients scale up fine, but do not scale down correctly. 
- Fractional can be a little ridiculous. 1 1/19 qt, 2 2/12 cups etc. For me is an easy decision, for other users it might prove confusing. Decimal might make it easier. Will wait on feedback. Or we can possibly round up or down.


Gemfile
```diff
@@ -4,26 +4,74 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby "3.3.7"
 
 gem "administrate"
```
nfoos, line 6: I'm a little confused why this is added here. Looks like you didn't branch from the latest `main`

app/classes/ingredient_calculator.rb
```diff
@@ -0,0 +1,77 @@
+class IngredientCalculator
+    attr_accessor :recipe
```
nfoos: Is there an `initialize` method that went missing? Have you been setting the recipe on the calculator after creating it?

app/classes/ingredient_calculator.rb
```diff
@@ -0,0 +1,77 @@
...
+
+        ingredients_text = recipe.ingredients.to_s
+        ingredients = ingredients_text.split(/r?\n/).map(&:strip).reject(&:empty?)
+
+        ingredients.map do |ingredient|
+            parse_ingredients(ingredient)
+        end
+    end
+
+    def parse_ingredients(ingredient)
```
nfoos: You should try wiring this up to the [recipe partial](https://github.com/CodeSmitty/mame-zefu/blob/4b2e5937f257f9c8b216040f61caac3f1a24ea48/app/views/recipes/_recipe.html.erb#L60) for the show view

Add a recipe helper method that uses this class to parse the ingredient, then wrap the unit and quantity in spans with a css class that you can verify at a glance, like make one red and the other blue. Then you can manually browse through various real recipes and see how it's performing

nfoos: You should also take note of any edge cases that you have to account for and add a unit test for them so that as you continue to refactor this you don't introduce regressions

app/classes/ingredient_parsing.rb
nfoos, line 1: These classes should be relocated to the `lib` directory, probably within the `recipes` namespace.  I'd also rename this one to parser instead of parsing. So this file would be `lib/recipes/ingredient_parser.rb` and the class would be `Recipes::IngredientParser`

app/classes/ingredient_parsing.rb
```diff
@@ -0,0 +1,50 @@
+require 'ingreedy'
+require_relative 'ingredient_constants'
+
+class IngredientParsing
+  include IngredientConstants
+
```
nfoos: I'd expect to see an initialize method in this class that takes a recipe as an argument then sets it to the @recipe class variable.

app/classes/ingredient_parsing.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def normalize_fractions(text)
+    normalized = text.dup
+    FRACTION_MAP.each do |unicode, replacement|
+      normalized.gsub!(unicode, replacement)
+    end
+    normalized
+  end
+end
```
nfoos: VSCode has some settings that will help with the final newline characters

Image

app/classes/ingredient_parsing.rb
```diff
@@ -0,0 +1,50 @@
...
+        ingredient: ingredient
+      }
+    end
+  end
+
+  def normalize_fractions(text)
+    normalized = text.dup
+    FRACTION_MAP.each do |unicode, replacement|
+      normalized.gsub!(unicode, replacement)
+    end
```
nfoos: I suggest looking into a more efficient way of doing this. There should be a way to not have to loop over every fraction in the map and perform that many substitutions every time

app/classes/ingredient_scaler.rb
```diff
@@ -0,0 +1,58 @@
...
+      scale_with_logic(parsed_ingredient, multiplier)
+    else
+      dont_scale(parsed_ingredient)
+    end
+  end
+
+  private
+
+  def should_scale?(ingredient)
+    !DO_NOT_SCALE.any? { |skip| ingredient.include?(skip) }
```
nfoos: This is very difficult to comprehend, I suggest rewording.

Gemfile
nfoos, line 1: ingreedy dependency is missing

